PR #40, opened 2026-02-25
Ac/psi 1548 1544

acostarelli:
Equivalent of PSI PRs #1544 and #1548


luke-kiernan (changes requested):
Two testing-related things (plus the above formatting).

acostarelli:
Yes my tests pass locally

PR #42, opened 2026-02-26
Ac/performance cats

acostarelli:


test/performance/performance_test_cats.jl
```diff
@@ -0,0 +1,84 @@
...
+    if length(ARGS) == 0 && !is_running_on_ci()
+        push!(ARGS, "Local Test")
+    end
+    write(io, "| $(ARGS[1]) | $(precompile_time.time) |\n")
+end
+
+try
+    # TODO Should probably be passed through argv but I think Github CI uses 
+    # argv already somehow?
+    cats_dir = "/Users/acostare/Documents/CATS-CaliforniaTestSystem/"
```
luke-kiernan: There's a way to tell GitHub CI actions "clone this repo locally as setup." But then it'd also have to read the time series data from the CSV, which takes ~15 minutes. Not good.

A more robust solution would be to add the CATS system to PSB's system catalog, so that we can use it across the Sienna ecosystem via `using PSB; sys = PSB.build_system(FooSystemCategory, "CATS_California")`. There was discussion of doing that a while back...alternatively, we convert the time series CSV into something cross-platform (is JDL2 portable between machines?) and upload it somewhere.

luke-kiernan:
Oh there's also one comment that got lost somehow: I see that you're having issues with the CATS system solving successfully. Check the device formulations and make sure they're reasonable, some sort of `Commitment` or `FixedOutput`, similar to what we're using in the other performance script for the UC step.

test/performance/performance_test_cats.jl
jd-lara: CATS is too large to try to run in the free CI-CD. This is a performance test geared towards offline testing IMO

jd-lara (approved):

Issue #51, opened 2025-02-25
Implement G-1 model


jd-lara:
Currently the models for SCUC are built for N - 1 transmission line outages and do not consider generator outages. 

There is a need to implement the SC model for generator outages

PR #62, opened 2026-03-16
move templates

jd-lara:


luke-kiernan:
Note to self: after [IOM #51](https://github.com/NREL-Sienna/InfrastructureOptimizationModels.jl/pull/51) gets resolved, we'll need to add `struct ServiceRequirementVariable <: VariableType end` around line 100 of `variables.jl`. But that's it: with that one change, the tests pass locally with IOM on the "decouple cost functions" branch. edit: okay done.

PR #63, opened 2026-03-20
update for pwl fcn name change in IOM

luke-kiernan:
POM equivalent of IOM PR [#55](https://github.com/NREL-Sienna/InfrastructureOptimizationModels.jl/pull/55)


luke-kiernan:
I fully expect the tests to fail because it's going to use the upstream main, not the upstream branch in the PR I've linked above.

Issue #43, opened 2026-02-27
Consolidate approach to models with RHS time series.


jd-lara:
In the current code base we have mixed methods on how time series are added to a model to constraint on the right hand side for ThermalStandard (derating), Interfaces, Interchanges, etc. Loads have yet another approach. 

After a refactor for DLRs, the approach should be consolidated as implemented here 

https://github.com/NREL-Sienna/PowerSimulations.jl/pull/1553

Roughly the approach is 

- Add a time series parameter if the attribute is added to the device model 
- when adding the Ub/lb constraints or the bounds then call a method to add the constraints with floats or with parameters. 

This will avoid unnecessary duplications of constraints. 

This is a large undertaking so we will split the work between @luke-kiernan and me.

luke-kiernan:
Clarifications:

1. Does this mean we're removing `ActivePowerVariableTimeSeriesLimitsConstraint` and instead just collapsing the keys to `ActivePowerVariableLimitsConstraint`?
2. For commitment formulations, we currently have two separate upper bounds `p <= pmax * forecast` and `p <= pmax * u`. Do you mean that we should combine them into one upper bound, `p <= pmax * forecast * u`? This will require new functions in IOM.

Currently, upper and lower bounds for semicontinuous are handled together in IOM. However, the presence of a time series only affect the upper bound, not the lower bound. If the answer to (2) is yes, I'll also need to go decouple those.

PR #72, opened 2026-04-09
Lk/issue 43 loads gate

luke-kiernan:
Address the issue raised in #43 for loads.


luke-kiernan:
I'm not sure if I should merge into `main` or into `lk/pom-tests`. The only difference between the two is one PR fixing some environment stuff, so it probably doesn't make a big difference one way or the other.

jd-lara (approved):

PR #75, opened 2026-04-14
Serialize systems to HDF

acostarelli:
@jd-lara Since there's no `HdfSimulationStore` equivalent in POM, the HDF file is just floating right now. Should a `system_hdf` attribute be added to EM/DM stores?


luke-kiernan (changes requested):
POM is single time step, single decision problem. This option is only relevant for simulation, multi-timestep. So just throw "oops not supported" if true and we'll address later.

jd-lara (approved):

PR #84, opened 2026-04-16
Add hydro turbine linear commitment formulation

acostarelli:
Port of https://github.com/NREL-Sienna/HydroPowerSimulations.jl/pull/126


src/common_models/add_parameters.jl
```diff
@@ -104,7 +104,7 @@ function _check_dynamic_branch_rating_ts(
         end
     end
 
-    multiplier = get_multiplier_value(T, device, W)
+    multiplier = get_multiplier_value(T(), device, W())
```
luke-kiernan: Original was correct. We pass types here, while PSI may still pass instances.
```suggestion
    multiplier = get_multiplier_value(T, device, W)
```

luke-kiernan (commented):
Based on the fact that you're passing types into `get_multiplier_value`, I'm inferring that this code wasn't actually run. Will review again once tests work

src/common_models/add_parameters.jl
acostarelli: The code was indeed run. This line did not cause any errors.

acostarelli: Confirmed, this line isn't run in testing.

acostarelli: I have a coverage PR coming up and I'll make sure this is covered.

src/static_injector_models/hydro_generation.jl
```diff
@@ -2230,29 +2245,44 @@ is_time_variant_term(
...
-            cost_term *= multiplier
-            exp = if d isa PSY.HydroPumpTurbine && PSY.get_must_run(d)
-                cost_term  # note we do not add this to the objective function
-            else
-                _add_proportional_term_maybe_variant!(
-                    Val(add_as_time_variant), container, U, d, cost_term, t)
-            end
-            add_to_expression!(container, ProductionCostExpression, exp, d, t)
-        end
+        cost_term = proportional_cost(op_cost_data, U, d, V)
```
acostarelli: @luke-kiernan Is this proper use of the new IOM code? Or would it be better to just have the one add_proportional_cost function? I don't know how much performance is lost to the "maybe variant" part.

luke-kiernan: Your definition here is non-time-varying, so I suspect it's redundant: the default `add_proportional_cost!` in IOM should handle it just fine.

luke-kiernan (commented):
There's bigger structural issues here: this file is still basically a copy-paste from PSI. In the initial PSI-into-POM/IOM split, I deferred refactoring the hydro code to "later," perhaps due to the separate HPS package. I never returned to that task. I've opened #96 for this and will tackle it ASAP (today and tomorrow).

Until that's finished, imo it doesn't make sense to merge this. Core issues first then add new features.

Aside: clearly we need better test coverage. This file calls functions that don't exist.

src/static_injector_models/hydrogeneration_constructor.jl
```diff
@@ -1801,6 +1801,17 @@ function construct_device!(
     return
 end
 
+_maybe_add_on_variables!(
```
acostarelli, line 1804: @luke-kiernan What do you think of this pattern?

src/core/formulations.jl
```diff
@@ -353,6 +359,15 @@ Formulation type to add injection variables for a HydroPumpTurbine only using en
 """
 struct HydroPumpEnergyCommitment <: AbstractHydroPumpFormulation end
 
+"""
```
acostarelli, line 362: @luke-kiernan and this? For simplifying unions

src/static_injector_models/hydrogeneration_constructor.jl
rodrigomha: I'm fine with this. @jd-lara What do you think?

src/core/formulations.jl
rodrigomha: I'm okay with this, since we have abstract defined for Commitments and Dispatch so this is okay

rodrigomha (approved):

luke-kiernan (approved):
If tests pass, I'm good to merge.

PR #85, opened 2026-04-16
Add shiftable load model

acostarelli:
Port of https://github.com/NREL-Sienna/PowerSimulations.jl/pull/1592


src/core/parameters.jl
```diff
@@ -238,6 +248,8 @@ convert_output_to_natural_units(::Type{ReactivePowerTimeSeriesParameter}) = true
 convert_output_to_natural_units(::Type{RequirementTimeSeriesParameter}) = true
 convert_output_to_natural_units(::Type{UpperBoundValueParameter}) = true
 convert_output_to_natural_units(::Type{LowerBoundValueParameter}) = true
+convert_result_to_natural_units(::Type{ShiftUpActivePowerTimeSeriesParameter}) = true
+convert_result_to_natural_units(::Type{ShiftDownActivePowerTimeSeriesParameter}) = true
```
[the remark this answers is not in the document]
luke-kiernan: CoPilot seems correct here: somehow the result -> output name change got missed.

src/core/variables.jl
```diff
@@ -124,6 +124,22 @@ Docs abbreviation: ``\\theta``
 """
 struct VoltageAngle <: VariableType end
 
+#########################################
+###### Power Load Shift Variables #######
+#########################################
+
+"""
+Struct to dispatch the creation of Shifted Active Power Variables
```
luke-kiernan:
```suggestion
Struct to dispatch the creation of Shifted Up Active Power Variables
```

src/core/variables.jl
```diff
@@ -124,6 +124,22 @@ Docs abbreviation: ``\\theta``
...
+Docs abbreviation: ``p^\\text{shift,up}``
+"""
+struct ShiftUpActivePowerVariable <: VariableType end
```
luke-kiernan, line 135: Might be worth introducing a common supertype: e.g. `abstract type ShiftPowerVariable <: VariableType end` then make the 2 types subtype `ShiftPowerVariableType` instead of `VariableType`.

src/static_injector_models/electric_loads.jl
```diff
@@ -24,6 +24,21 @@ get_multiplier_value(::Type{<:TimeSeriesParameter}, d::PSY.ElectricLoad, ::Type{
 get_multiplier_value(::Type{ReactivePowerTimeSeriesParameter}, d::PSY.ElectricLoad, ::Type{StaticPowerLoad}) = -1*PSY.get_max_reactive_power(d)
 get_multiplier_value(::Type{<:TimeSeriesParameter}, d::PSY.ElectricLoad, ::Type{<:AbstractControllablePowerLoadFormulation}) = PSY.get_max_active_power(d)
 
+########################### ShiftablePowerLoad #####################################
+
+get_variable_binary(::Type{ShiftUpActivePowerVariable}, ::Type{<:PSY.ElectricLoad}, ::Type{PowerLoadShift}) = false
```
luke-kiernan, line 29: Identical definitions for `Up` and `Down`: if you have a common supertype, you could eliminate the duplication and just write one definition, for `Type{<:ShiftPowerVariable}`.

src/static_injector_models/electric_loads.jl
```diff
@@ -180,6 +310,154 @@ function add_constraints!(
...
+        T,
+        V,
+        PSY.get_name.(devices),
+    )
+    up_variable = get_variable(container, ShiftUpActivePowerVariable, V)
+    down_variable = get_variable(container, ShiftDownActivePowerVariable, V)
+    jump_model = get_jump_model(container)
+    for d in devices
+        name = PSY.get_name(d)
+        constraint[name] =
```
luke-kiernan: This is a 1D constraint container, when most things are 2D. iirc those things have caused bugs and headaches in the past. I'd at least add a `FIXME` comment citing a relevant issue (pretty sure there is one somewhere).

src/static_injector_models/electric_loads.jl
```diff
@@ -180,6 +310,154 @@ function add_constraints!(
...
+            JuMP.@constraint(
+                jump_model,
+                sum(up_variable[name, t] - down_variable[name, t] for t in time_steps) ==
+                0.0
+            )
+    end
+    additional_balance_interval = get_attribute(model, "additional_balance_interval")
+    if !isnothing(additional_balance_interval)
```
luke-kiernan, line 343: If not nothing, what's the expected type of `additional_balance_interval`? What are acceptable values, and where do we perform input validation?

My concern: what's stopping it from being the wrong type, not a multiple of `resolution`, or negative?

src/static_injector_models/electric_loads.jl
```diff
@@ -184,14 +310,162 @@ function add_constraints!(
...
+            Dates.Millisecond(resolution).value
+        for d in devices
+            name = PSY.get_name(d)
+            constraint_aux[name] = JuMP.@constraint(
+                container.JuMPmodel,
+                sum(
+                    up_variable[name, t] - down_variable[name, t] for
+                    t in 1:interval_length
+                ) == 0.0
+            )
```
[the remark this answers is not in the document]
luke-kiernan: Yeah reasonable questions about correctness: is enforcing sum zero over all t's in `[1, additional_balance_interval // period]` the desired behavior?

src/static_injector_models/electric_loads.jl
```diff
@@ -34,6 +49,8 @@ proportional_cost(cost::Nothing, ::Type{OnVariable}, ::PSY.ElectricLoad, ::Type{
 proportional_cost(cost::PSY.OperationalCost, ::Type{OnVariable}, ::PSY.ElectricLoad, ::Type{<:AbstractControllablePowerLoadFormulation})=PSY.get_fixed(cost)
 
 objective_function_multiplier(::Type{<:VariableType}, ::Type{<:AbstractControllablePowerLoadFormulation})=OBJECTIVE_FUNCTION_NEGATIVE
+objective_function_multiplier(::Type{ShiftUpActivePowerVariable}, ::Type{<:AbstractControllablePowerLoadFormulation})=OBJECTIVE_FUNCTION_NEGATIVE
```
luke-kiernan: Mismatch here: first definition is for any subtype of `AbstractControllablePowerLoadFormulation`, second for only `PowerLoadShift`. Purposeful?

src/static_injector_models/electric_loads.jl
```diff
@@ -267,3 +545,30 @@ function proportional_cost(
...
+    devices::IS.FlattenIteratorWrapper{T},
+    ::DeviceModel{T, U},
+    ::Type{<:PM.AbstractPowerModel},
+) where {T <: PSY.ShiftablePowerLoad, U <: PowerLoadShift}
+    add_variable_cost!(container, ShiftUpActivePowerVariable, devices, U)
+    add_variable_cost!(container, ShiftDownActivePowerVariable, devices, U)
+    return
+end
+
+### Special Method to skip VOM cost on ShiftUpActivePowerVariable ###
```
luke-kiernan, line 574: Why `Up` only, not `Down`?

luke-kiernan (changes requested):
Big-picture question: why do we need both `Up` and `Down`? I believe we could do everything here with one variable, `net = up - down`:

- bounds: `0 <= up <= max_up(t)` and `0 <= down <= max_down(t)` are equivalent to `-max_down(t) <= net <= max_up(t)`
- constraints: all in terms of `up-down` or `down-up`.
- objective: up and down use the same variable cost with opposite signs. So you get `-c(t) * up + c(t) * down` which is just `-c(t) * net`.

Who opened the corresponding PSI PR? If they've thought of all these things and addressed them, then I'm okay to merge--just want to check.

src/core/parameters.jl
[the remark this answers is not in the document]
rodrigomha: Yeah, let's apply the suggestion.

src/core/variables.jl
rodrigomha: I think with two options is fine to use Unions. The moment we have three is okay to add another abstract probably. @jd-lara What's your opinion here?

src/static_injector_models/electric_loads.jl
rodrigomha: Ditto here. @jd-lara should we add the abstract?

src/static_injector_models/electric_loads.jl
rodrigomha: I think is good idea to be specific. Just concern about ambiguities if we change it.

src/static_injector_models/electric_loads.jl
rodrigomha: Agree here. 

@copilot Can you fix this container to add another dimension, such that the second dimension is a vector with a single element of timesteps[end].

src/static_injector_models/electric_loads.jl
rodrigomha: This is a discussion on how to add cost. We expect that the ShiftDown is the only one that introduce operational cost via VOM, while ShiftUp is free to use it.

src/static_injector_models/electric_loads.jl
rodrigomha: It's the responsability of the user adding the case to specify the Hour(X) such that an additional balance is useful. If they include a input we should fail here. We could do the interval_length a previous check to give a better error.

rodrigomha (changes requested):
We can wait on the discussion of the abstract. But overall is good.

src/core/variables.jl
luke-kiernan: If you considered it and decided against it, that's fine--code duplication is pretty minimal. Will mark as resolved.

src/static_injector_models/electric_loads.jl
luke-kiernan: Just to make sure we're on the same page: `Up` has 2nd argument as any subtype of `AbstractControllablePowerLoadFormulation`, `Down` has just `PowerLoadShift`. That mismatch is purposeful? I'd expect `PowerLoadShift` for both

rodrigomha: Yeah, we should change it for both cases using PowerLoadShift.

src/services_models/agc.jl
```diff
@@ -141,16 +141,18 @@ function add_constraints!(
...
+    R_up = get_variable(container, DeltaActivePowerUpVariable, U)
+    R_dn = get_variable(container, DeltaActivePowerDownVariable, U)
+    # FIXME dead code, file is unincluded, but likely bug: UpVariable for both.
+    # PSI issue #1593.
     R_up_emergency =
-        get_variable(container, AdditionalDeltaActivePowerUpVariable(), PSY.Area)
+        get_variable(container, AdditionalDeltaActivePowerUpVariable, PSY.Area)
     R_dn_emergency =
-        get_variable(container, AdditionalDeltaActivePowerUpVariable(), PSY.Area)
+        get_variable(container, AdditionalDeltaActivePowerUpVariable, PSY.Area)
```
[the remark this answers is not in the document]
luke-kiernan: Known issue: PSI [#1593](https://github.com/Sienna-Platform/PowerSimulations.jl/issues/1593). File isn't included, so not really a major problem.

PR #87, opened 2026-04-17
Add hydro formulation with bilinear approximation

acostarelli:
Requires IOM on : https://github.com/Sienna-Platform/InfrastructureOptimizationModels.jl/pull/86


src/static_injector_models/hydro_generation.jl
```diff
@@ -1802,6 +1806,81 @@ function add_constraints!(
...
+            V,
+            PSY.get_name.(reservoirs),
+            time_steps,
+            flow[name, :, :],
+            head,
+            [
+                (
+                    min = get_variable_lower_bound(HydroTurbineFlowRateVariable, d, W),
+                    max = get_variable_upper_bound(HydroTurbineFlowRateVariable, d, W),
+                ) for _ in 1:length(reservoirs)
```
luke-kiernan: Oh, you're passing in the same bounds for all the reservoirs here. You could make that clearer via `repeat`.

It took me a bit of thought to work out the shapes: might be worth mentioning that we're assuming min/max bounds are time-independent here.

luke-kiernan (changes requested):
More so than the below comment, I'd like to see one or two unit tests. IOM tests the formulation math, so you don't need to do that. The concern is more so: does the POM plumbing hand things to IOM correctly? E.g. spot check a couple constraints and coefficients for a simple case, to confirm that things are getting matched up as intended, that we're not missing a unit conversion somewhere

src/static_injector_models/hydro_generation.jl
acostarelli: Do min/max bounds change over time in other cases?

luke-kiernan: Time-varying `max_active_power` is super common, e.g. for solar or wind. I can't find a spot where head height or outflow limits are time-varying, though.

test/test_device_hydro_constructors.jl
```diff
@@ -692,6 +692,183 @@ end
...
+
+    total_inflow = sum(values(hydro_inflow_ts))
+    total_outflow = sum(df_outflow[!, :value])
+    total_spillage = sum(hydro_spillage_df[!, :value])
+
+    calculated_vf =
+        (hydro_vol_df[1, :value]) +
+        ((total_inflow - total_outflow - total_spillage) * 3600 * 1e-9)
+
+    @test abs(calculated_vf - hydro_vol_df[end, :value]) <= 1e-4
```
luke-kiernan: This looks like a good use case for `approx`. Might be worth commenting why 1e-4 is reasonable: presumably `calculated_vf` is signficiantly larger? Some magic numbers in `calculate_vf` too.

src/static_injector_models/hydro_generation.jl
```diff
@@ -1791,6 +1791,82 @@ function add_constraints!(
...
+            container,
+            V,
+            PSY.get_name.(reservoirs),
+            time_steps,
+            flow[name, :, :],
+            head,
+            repeat(
+                [
+                    (
+                    min = get_variable_lower_bound(HydroTurbineFlowRateVariable, d, W),
```
luke-kiernan, line 1866: nitpick: might be worth defining a function for this `(variable, device, formulation_type) -> (min = ..., max = ...)` pattern. Only 2 spots right now (so probably not), but will new PRs add more?

luke-kiernan (approved):
My other 2 cents (more of a nitpick): the commit situation is 10x more complex than the code merits...I'd be tempted to merge from a fresh branch with just 1 or 2 commits.

test/test_device_hydro_constructors.jl
acostarelli: I copied this test from the other Bilinear formulation tbh

rodrigomha: I agree with Luke here. I would replace the test by:
`@test isapprox(calculated_vf, hydro_vol_df[end, :value], atol = tol`

In the same sense, can you define `tol = 1e-4` and a comment to why using 1e-4?

test/test_device_hydro_constructors.jl
```diff
@@ -692,6 +692,183 @@ end
...
+    hydro_inflow_df =
+        read_parameters(outputs, [(InflowTimeSeriesParameter, HydroReservoir)])["InflowTimeSeriesParameter__HydroReservoir"]
+
+    total_inflow = sum(values(hydro_inflow_ts))
+    total_outflow = sum(df_outflow[!, :value])
+    total_spillage = sum(hydro_spillage_df[!, :value])
+
+    calculated_vf =
+        (hydro_vol_df[1, :value]) +
+        ((total_inflow - total_outflow - total_spillage) * 3600 * 1e-9)
```
rodrigomha: Rather than using 3600 and 1e-9, can you use POM.SECONDS_IN_HOUR and POM.M3_TO_KM3, as defined here: https://github.com/Sienna-Platform/PowerOperationsModels.jl/blob/main/src/core/physical_constant_definitions.jl

rodrigomha (changes requested):
Just minor comments in the tests for readability and trying to be consistent. After those changes we should be ready to merge.

rodrigomha (approved):
this is looking good

jd-lara (approved):

PR #77, opened 2026-04-14
Update POM for PSY MarketBidCost/ImportExportCost static/TS split

luke-kiernan:
## Summary

PSY split `MarketBidCost` and `ImportExportCost` into separate static and time-series variants (`MarketBidCost` + `MarketBidTimeSeriesCost`, `ImportExportCost` + `ImportExportTimeSeriesCost`). This PR mirrors the IOM refactor ([IOM PR #74](https://github.com/NREL-Sienna/InfrastructureOptimizationModels.jl/pull/74)) on the POM side and adds a unit-test suite for the MBC/IEC objective-function construction.

Depends on IOM PR [#74](https://github.com/Sienna-Platform/InfrastructureOptimizationModels.jl/pull/88).

### Dispatch (src/)

- **Consolidated MBC `OnVariable` dispatch** into a single generic pair in `common_models/market_bid_overrides.jl`, keyed on an `_onvar_offer_direction` trait (`Generator → IncrementalOffer`, `ControllableLoad → DecrementalOffer`). Per-device thermal/hydro/load overrides drop out.
- **Replaced stale `isnothing(get_{input,output}_offer_curves(...))` checks** with `IOM.is_nontrivial_offer` — `ZERO_OFFER_CURVE` is now the default, not `nothing`. Load dispatch rejects non-trivial supply with `ArgumentError`.
- **Delegated hydro `add_proportional_cost!`** to IOM's shared `add_proportional_cost_maybe_time_variant!`, dropping a broken `_lookup_maybe_time_variant_param` path that never had a definition.
- **Collapsed `is_time_variant_term` to a one-arg form** (output depends only on cost type).
- **`_get_time_series_name` for `*CostAtMinParameter`** traverses through the curve — `initial_input` now lives inside `TimeSeriesPiecewiseIncrementalCurve`, not on MBC directly.

### Tests

Existing MBC/IEC helpers (`add_market_bid_cost.jl`, `mbc_system_utils.jl`, `iec_simulation_utils.jl`) updated for `LinearCurve` scalar fields and the new TS cost constructors.

New unit-test suite, focused on "do the objective function terms and constraints turn out correct":

[...]

PR #95, opened 2026-04-22
Increase code coverage

acostarelli:
This PR is based off #77

### DIFF
--- a/src/energy_storage_models/storage_models.jl
+++ b/src/energy_storage_models/storage_models.jl
@@ -1425,7 +1425,7 @@ function add_cycling_charge_without_reserves!(
             PSY.get_conversion_factor(d)
         cycle_count = PSY.get_cycle_limits(d)
         efficiency = PSY.get_efficiency(d)
-        constraint[name, time_steps[end]] = JuMP.@constraint(
+        constraint[name] = JuMP.@constraint(
             get_jump_model(container),
             sum((
                 powerin_var[name, t] * efficiency.in * fraction_of_hour for t in time_steps
@@ -1435,7 +1435,6 @@ function add_cycling_charge_without_reserves!(
     return
 end
 
-# no test coverage
 function add_cycling_charge_with_reserves!(
     container::OptimizationContainer,
     devices::IS.FlattenIteratorWrapper{V},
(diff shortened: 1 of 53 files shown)


src/energy_storage_models/storage_models.jl
```diff
@@ -1425,7 +1425,7 @@ function add_cycling_charge_without_reserves!(
             PSY.get_conversion_factor(d)
         cycle_count = PSY.get_cycle_limits(d)
         efficiency = PSY.get_efficiency(d)
-        constraint[name, time_steps[end]] = JuMP.@constraint(
+        constraint[name] = JuMP.@constraint(
```
luke-kiernan: Looks like you're a commit or 2 behind from `lk/mbc_iec_refactor`--before you merge, take a careful look at the non-test changes and make sure they're intentional (not a reversion to fixes). Rebasing might do the trick too.

luke-kiernan (changes requested):
One small regression in the code body. But otherwise good! Evidently it's helping--you caught a couple silly typos

src/energy_storage_models/storage_models.jl
acostarelli: Okay I'll double-check. This particular change was intentional, the constraint container is only instantiated with one axis.

luke-kiernan: Oh did CoPilot change the indexing [here](https://github.com/Sienna-Platform/PowerOperationsModels.jl/pull/85#discussion_r3120398425) without actually doing anything to make the underlying container 2D?

test/test_transfer_initial_conditions.jl
```diff
@@ -33,3 +33,83 @@
     # ED should still solve with the transferred ICs
     @test solve!(ed) == IOM.RunStatus.SUCCESSFULLY_FINALIZED
 end
+
+@testset "transfer_initial_conditions! with EnergyReservoirStorage" begin
```
luke-kiernan, line 37: Just to clarify: I wrote `transfer_initial_conditions` because I needed its functionality for the performance script. It's a mildly hacky way to do multi timestep things within POM, not really official interface. (Still good to have tests for it, though.)

luke-kiernan (changes requested):
The 1D vs 2D container question is worth looking into.

Keep in mind that we should also test numerical correctness, not just "does it build and solve." However, let's defer that to a later PR: writing those tests may be a significant undertaking.

src/energy_storage_models/storage_models.jl
acostarelli: It looks like Copilot did it correctly there on electric loads.

PR #103, opened 2026-04-29
Update MT-HVDC quadratic converter model

acostarelli:


src/mt_hvdc_models/hvdcsystems_constructor.jl
```diff
@@ -156,75 +76,89 @@ function construct_device!(
     container::OptimizationContainer,
     sys::PSY.System,
     ::ModelConstructStage,
-    model::DeviceModel{PSY.InterconnectingConverter, QuadraticLossConverter},
+    model::DeviceModel{PSY.InterconnectingConverter, Bin2QuadraticLossConverter},
```
acostarelli: There's a world where we combine the two model construct stages too, but this would force us to calculate the V^2 expression for the NLP version which it doesn't need, although it may not affect the solver if it doesn't appear in any constraints or the objective.

test/test_device_hvdc.jl
```diff
@@ -127,3 +121,28 @@ end
           IOM.ModelBuildStatus.BUILT
     @test solve!(model) == IOM.RunStatus.SUCCESSFULLY_FINALIZED
 end
+
+@testset "HVDC System with Losses Network (QuadraticLossConverter NLP)" begin
```
rodrigomha: We should add a test that have both together and check that they are in the same order of magnitude for current and dc power probably.

You can run the same system and run without linear loss for both.

src/mt_hvdc_models/HVDCsystems.jl
```diff
@@ -564,197 +541,86 @@ function add_constraints!(
...
-        )
-
-    use_linear_loss = get_attribute(model, "use_linear_loss")
-    if use_linear_loss
-        pos_current = get_variable(container, ConverterPositiveCurrent, U)
-        neg_current = get_variable(container, ConverterNegativeCurrent, U)
+    P_ac_var = get_variable(container, ActivePowerVariable, U)
+    vi_expr = get_expression(container, IOM.BilinearProductExpression, U, "vi")
+    i_sq_expr = get_expression(container, IOM.QuadraticExpression, U, "i_sq")
+    if get_attribute(model, "use_linear_loss")
```
luke-kiernan: This is type-unstable...but if this is a runtime setting, then there's no avoiding it. (If we're always using the default attributes then we should switch to a multiple dispatch pattern.)

src/common_models/add_to_expression.jl
```diff
@@ -718,6 +718,45 @@ function add_to_expression!(
...
+    W <: AbstractTwoTerminalVSCFormulation,
+    X <: ACPPowerModel,
+}
+    var = get_variable(container, U, V)
+    nodal_expr = get_expression(container, T, PSY.ACBus)
+    network_reduction = get_network_reduction(network_model)
+    time_steps = get_time_steps(container)
+    for d in devices
+        name = PSY.get_name(d)
+        bus_no = PNM.get_mapped_bus_number(
```
luke-kiernan, line 748: I'm a little surprised to see this `get_mapped_bus_number` call here. What's the status of our support and testing for network reductions? Problem is that anything bus-associated gets weird...

src/core/variables.jl
```diff
@@ -150,108 +150,32 @@ Docs abbreviation: ``i_l^{dc}``
...
-
-"""
-Struct to dispatch the creation of Auxiliary Variable for Converter Bilinear term: v * i
-Docs abbreviation: ``\\gamma_c^{dc}``
+Positive part of an absolute-value current decomposition (``i = i^+ - i^-``).
+Used by both two-terminal HVDC links (cable current) and interconnecting converters.
+Docs abbreviation: ``i^{+,dc}``
 """
-struct AuxBilinearConverterVariable <: VariableType end
+struct PositiveCurrent <: VariableType end
```
luke-kiernan: Generic name for something with a pretty narrow use case. Call it `PositiveAbsDecompCurrent` instead? Oh but we like to use full words for public interface where possible...

src/core/variables.jl
```diff
@@ -150,108 +150,32 @@ Docs abbreviation: ``i_l^{dc}``
...
 
 """
-Struct to dispatch the creation of Auxiliary Variable for Squared Converter Bilinear term: v * i
-
-Docs abbreviation: ``\\gamma_c^{sq,dc}``
-"""
-struct AuxBilinearSquaredConverterVariable <: VariableType end
-
+Negative part of an absolute-value current decomposition (``i = i^+ - i^-``).
```
luke-kiernan: > Negative part of an absolute-value current decomposition (``i = i^+ - i^-``)

Is this "absolute value decomposition" thing standard terminology? I had to go pull up `_add_abs_value_decomposition_constraints!` in order to figure out what's going on. Namely, only one of `i^+` and `i^-` can be nonzero, and whichever one is nonzero equals `abs(i)`.

src/core/formulations.jl
```diff
@@ -201,9 +226,27 @@ Linear Loss InterconnectingConverter Model
...
-Quadratic Loss InterconnectingConverter Model
+Abstract supertype for InterconnectingConverter formulations with quadratic losses.
+"""
+abstract type AbstractQuadraticLossConverter <: AbstractConverterFormulation end
+
+"""
+Quadratic Loss InterconnectingConverter using the separable bilinear approximation
+(`v·i = ½((v+i)² − v² − i²)`) with a SOS2-based PWL approximation for x². Stays MILP.
+"""
+struct MILPQuadraticLossConverter <: AbstractQuadraticLossConverter end
```
luke-kiernan: nitpick: it'd be more consistent to put MILP at the end of the name, since we have `HVDCTwoTerminalVSCNLP`. Similarly I'd maybe add a `MINLP` to the end of `QuadraticLossConverter`.

src/mt_hvdc_models/HVDCsystems.jl
```diff
@@ -564,381 +543,58 @@ function add_constraints!(
...
-                    a * var_sq_current[name, t] +
-                    b * (pos_current[name, t] + neg_current[name, t]) + c
-            else
-                loss = a * var_sq_current[name, t] + c
-            end
-            constraint[name, t] = JuMP.@constraint(
-                get_jump_model(container),
-                var_ac_power[name, t] == var_dc_power[name, t] - loss
+            i_pos_t = use_linear_loss ? i_pos_var[name, t] : nothing
+            i_neg_t = use_linear_loss ? i_neg_var[name, t] : nothing
```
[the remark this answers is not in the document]
luke-kiernan: Good catch. Also add a test case for this

luke-kiernan (commented):
Took a brief look--a few small things. But you should have someone more familiar with these formulations also take a look

src/core/formulations.jl
acostarelli: Will be in later issue

src/mt_hvdc_models/HVDCsystems.jl
acostarelli: True

src/common_models/add_to_expression.jl
acostarelli: LCC does something similar.

rodrigomha: Is needed for network reductions I think

src/common_models/quadratic_converter_loss.jl
```diff
@@ -0,0 +1,101 @@
...
+    iszero(a) || JuMP.add_to_expression!(expr, a, i_sq_t)
+    iszero(b) || JuMP.add_to_expression!(expr, b, abs_i_t)
+    return expr
+end
+
+#########################################
+######## Absolute-value surrogate #######
+#########################################
+
+function _add_abs_value_variables!(
```
rodrigomha: Why do we need this method? Can we just call add_variables! directly in the constructor?

src/common_models/quadratic_converter_loss.jl
```diff
@@ -0,0 +1,101 @@
...
+    i_sq_t::JuMP.AffExpr,
+    abs_i_t,
+)
+    expr = JuMP.AffExpr(c)
+    iszero(a) || JuMP.add_to_expression!(expr, a, i_sq_t)
+    iszero(b) || JuMP.add_to_expression!(expr, b, abs_i_t)
+    return expr
+end
+
+function _quadratic_converter_loss_expr(
```
rodrigomha, line 44: This methods are good. We should add docstrings to explain here that we skip the terms.

Also, can we specify that abs_i_t is a JuMP something here so we avoid compiling an Any for abs_i_t?

rodrigomha (changes requested):
This is looking good. Just minor comments and we are good to go.

rodrigomha (approved):
I'm happy with this if tests are passing locally

jd-lara (approved):

PR #104, opened 2026-04-30
Port hybrid systems

acostarelli:
Thanks for opening a PR to PowerOperationsModels.jl, please take note of the following when making a PR:

Check the [contributor guidelines](https://sienna-platform.github.io/PowerOperationsModels.jl/stable/api/developer_guidelines/)


rodrigomha (changes requested):
We need to add a couple of extra tests.

First, we need to test different attributes, that is using reservation = false, and energy target = true. Also, what happened with storage_reservation and regularization attributes?

Also, it would be good to test that the hybrid system builds when there is no thermal, or no storage, or no renewable.

Finally, it would be good to test when the hybrid does not participate in reserves.

Overall, the PR looks good. Once we have the tests, I will QA/QC the model for the different tests to ensure that the constraints are looking good.

src/hybrid_system_models/hybrid_systems.jl
```diff
@@ -0,0 +1,1828 @@
+#! format: off
```
rodrigomha: I think we can enable format on every line here. The methods are quite large and probably need format anyways

src/core/parameters.jl
```diff
@@ -222,6 +222,18 @@ Parameter to record that the component changed in the availability status
 """
 struct AvailableStatusChangeCountdownParameter <: EventParameter end
 
+#################################################################################
+# Hybrid System Parameters
+#################################################################################
+
+"Time-series parameter for the maximum active power available from a hybrid system's
```
kdayday, line 229: @jd-lara This is a change from how things are currently defined in HSS with the time series attached to the hybrid, not the subcomponent, but I guess that works and is simpler for this case?

kdayday (changes requested):
Following Rodrigo's lead on the testing, but in addition, can you port over the updated equivalent of what's in the HybridDispatchWithReserves docstring on the HSS PR and have claude add hyperlinks to exported types and functions referenced in the docstring? It looks like the POM docs are way out of date, but would like to update as we go.

src/core/formulations.jl
```diff
@@ -409,14 +409,205 @@ abstract type AbstractHybridFormulation <: IOM.AbstractDeviceFormulation end
...
+attached the formulation collapses to an energy-only hybrid dispatch.
+
+Use with a hybrid system in a [`DeviceModel`](@ref) for unit commitment or economic
+dispatch.
+
+**Variables:**
+
+  - [`ActivePowerOutVariable`](@ref):
+
+      + Bounds: [0.0, ``P_{\\max,\\text{pcc}}``]
```
acostarelli: I think `domain` might be more accurate here

src/hybrid_system_models/hybrid_systems.jl
```diff
@@ -680,22 +586,24 @@ function add_to_expression!(
...
     X <: AbstractPowerModel,
 }
     expression = get_expression(container, T, V)
     time_steps = get_time_steps(container)
+    skip_kind = _excluded_reserve_kind(T)
     for d in devices
         name = PSY.get_name(d)
         for service in PSY.get_services(d)
-            isa(service, PSY.Reserve{PSY.ReserveUp}) && continue
+            service isa skip_kind && continue
```
acostarelli: Don't use `isa`. Add a method to handle this, or restructure existing dispatch.

src/hybrid_system_models/hybrid_systems.jl
```diff
@@ -1930,129 +1676,82 @@ function add_constraints!(
...
-        V,
-        names,
-        time_steps,
-    )
+    invert = _pcc_invert_reservation(T)
+    constraint = add_constraints_container!(container, T, V, names, time_steps)
 
     has_reserves = W <: AbstractHybridFormulationWithReserves && has_service_model(model)
-    r_up, r_dn = if has_reserves
+    r_ub, r_lb = if has_reserves
```
acostarelli: Combine these if statements.

src/core/formulations.jl
```diff
@@ -403,3 +403,249 @@ The formulation supports the following attributes when used in a [`PowerSimulati
...
+      + Domain: [0.0, ]
+      + Symbol: ``sb^{\\text{in}}_t``
+
+  - [`ChargeRegularizationVariable`](@ref), [`DischargeRegularizationVariable`](@ref)
+    (only when `"regularization" => true`): non-negative slacks bounding step changes in
+    charge/discharge between consecutive time steps.
+
+**Time Series Parameters:**
+
+| Parameter | Default Time Series Name |
```
kdayday, line 542: @acostarelli this documents how things are defined in the current HybridSystemSimulations. Looking at where `HybridRenewableActivePowerTimeSeriesParameter` is defined below, I'm actually not sure if in POM the time series is attached to the HybridSystem vs. the subcomponents, but I'm also not finding any other references to `"RenewableDispatch__max_active_power"` besides the comment, so I assume the time series are not fully implemented as they are in HSS?

rodrigomha:
Major concern here regarding parametric structs. Did we decide to go using Parametric types for defining constraints or expressions? I think it could be a good plan here for Hybrid and all the combinations rather than having multiple structs.

What are your thoughts here @jd-lara ? For example we have expressions with 3 parametric types, e.g. ReserveAggregationExpression.

My main comment is that we have a lot of const for handling the different parametric types. @acostarelli What is the reason for having multiple const rather than multiple structs? Does using parametric types makes it better to reuse code with multiple dispatch? I'm not following that much this, but could be?

In any case @acostarelli if we decide to go in this direction, we should change the generic Up and Down structs for something better defined, e.g. UpDirection and DownDirection, rather than simply Up and Down

I will use TotalReserve rather than UnscaledReserve.  You also have the const using Total for the Unscaled type.

Also, I'm not following exactly what ReserveSide with ChargeSide and DischargeSide. I think maybe eliminating this ReserveSide thing, and only leave two parametric types could be simpler than having so many consts.

Overall the PR looks quite good, so I would just focus on deciding about the parametric types, but the tests looks quite comprehensive.

src/core/reserve_traits.jl
```diff
@@ -0,0 +1,25 @@
+# Marker singleton trait types used to parametrize hybrid/storage reserve variable,
+# expression, and constraint families. These remove the need for paired sibling
+# singletons across the codebase: a single parametric struct + const aliases replaces
+# every (Charge/Discharge), (Up/Down), (Unscaled/Deployed), (UB/LB) sibling pair.
+
+abstract type ReserveDirection end
```
jd-lara: These look duplicates from PSY

src/core/reserve_traits.jl
```diff
@@ -0,0 +1,25 @@
...
+abstract type ReserveSide end
+"Discharge / outflow side of a storage or hybrid PCC. Was Out (PCC) / Discharge (storage)."
+struct DischargeSide <: ReserveSide end
+"Charge / inflow side of a storage or hybrid PCC. Was In (PCC) / Charge (storage)."
+struct ChargeSide <: ReserveSide end
+
+# Constraint UB/LB axis: reuse IOM's `BoundDirection` / `UpperBound` / `LowerBound`
+# (defined in InfrastructureOptimizationModels/common_models/constraint_helpers.jl).
+# A local alias keeps the abstract name discoverable through POM exports.
+const ConstraintBound = InfrastructureOptimizationModels.BoundDirection
```
jd-lara: This is duplicative

src/core/formulations.jl
acostarelli: I updated the docs to clarify that the time series are on the HS not the subcomponent. I'm not sure if I fully understand, but does this address your concern?

src/core/expressions.jl
```diff
@@ -78,25 +78,46 @@ of the energy balance for the system in medium term planning
...
-struct ReserveDeploymentBalanceUpDischarge <: StorageReserveDischargeExpression end
-struct ReserveDeploymentBalanceUpCharge <: StorageReserveChargeExpression end
-struct ReserveDeploymentBalanceDownDischarge <: StorageReserveDischargeExpression end
-struct ReserveDeploymentBalanceDownCharge <: StorageReserveChargeExpression end
+"""
+Aggregation of reserve variables allocated to the storage subcomponent of a hybrid system
+(or a standalone storage device).
+"""
+struct StorageReserveBalanceExpression{D, S, Sd} <:
+       ReserveAggregationExpression{D, S, Sd} end
```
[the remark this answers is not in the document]
acostarelli: We're not concerned with any API breaking.

src/energy_storage_models/storage_constructor.jl
```diff
@@ -9,14 +9,14 @@ function _add_ancillary_services!(
...
-        ReserveDeploymentBalanceDownCharge,
+        StorageReserveBalanceExpression{Up, UnscaledReserve, DischargeSide},
+        StorageReserveBalanceExpression{Up, UnscaledReserve, ChargeSide},
+        StorageReserveBalanceExpression{Down, UnscaledReserve, DischargeSide},
+        StorageReserveBalanceExpression{Down, UnscaledReserve, ChargeSide},
+        StorageReserveBalanceExpression{Up, DeployedReserve, DischargeSide},
+        StorageReserveBalanceExpression{Up, DeployedReserve, ChargeSide},
+        StorageReserveBalanceExpression{Down, DeployedReserve, DischargeSide},
+        StorageReserveBalanceExpression{Down, DeployedReserve, ChargeSide},
     ]
```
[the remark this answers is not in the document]
acostarelli: Maybe it's better we don't touch storage systems since that's out of scope for this PR, and then we can move the reserve_traits.jl to just live in hybrid_systems.jl so the impacted area is less.

src/hybrid_system_models/hybrid_systems.jl
```diff
@@ -0,0 +1,2389 @@
...
+            )
+        end
+    else  # EndOfPeriod
+        for t in time_steps
+            con[ci_name, t] = JuMP.@constraint(
+                jm,
+                sustained_param_discharge * reserve_var[ci_name, t] <=
+                energy_var[ci_name, t]
+            )
+        end
```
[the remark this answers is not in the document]
acostarelli: @rodrigomha @jd-lara What do you think of this? I accepted the change, but this isn't how HSS does it.

PR #105, opened 2026-05-04
implement native networkmodels

jd-lara:
This is part 1 of removing the dependency on PowerModels. More testing ideas from what's here already are welcome.

## Update: rebase onto main + PSY6/IOM-main units management

Per @jd-lara's request, the branch was updated to current `main` with the new dependencies and unit-system management:

- Merged `origin/main` into the branch and resolved conflicts in `src/ac_transmission_models/branch_constructor.jl` (kept both the new `Transformer3W` native DCP/ACP `construct_device!` methods and the new `TwoTerminalVSCLine` methods from main) and `test/Project.toml`.
- Pointed `[sources]` at the Sienna-Platform org with `IS4`, `psy6`, and `main` for `InfrastructureOptimizationModels` (plus `psy6` for `PowerNetworkMatrices`/`PowerSystemCaseBuilder` and `lk/psy6-units` for `PowerFlows` in the test environment).
- Renamed leftover `OperationsProblemTemplate` references in the native DCP/ACP `construct_network!` methods to `PowerOperationsProblemTemplate`.
- Added the `PSY.SU` unit-system argument to PSY accessors used by the native `branch_admittance` and `branch_flow_limits` helpers (`get_series_admittance`, `get_b`, `get_primary_shunt`, `get_r`, `get_x`, `get_flow_limits`, `get_rating`) to align with PSY6 conventions.
- Updated `test/test_native_dcp_acp_models.jl` to use the unit-system aware accessors.

`test_native_dcp_acp_models` and `test_network_constructors` pass under PSY6/IOM-main. Some unrelated pre-existing variable-bound accessors in `src/ac_transmission_models/AC_branches.jl` (the `get_variable_upper_bound`/`get_variable_lower_bound` definitions for `MonitoredLine`, `TapTransformer`, and `Transformer2W`) still call `PSY.get_flow_limits(d)`/`PSY.get_rating(d)` without `PSY.SU`; these surface in `test_device_branch_constructors` after the PSY6 bump but are outside the native NetworkModel scope of this PR.


src/network_models/acp_model.jl
```diff
@@ -0,0 +1,166 @@
+# Native ACP network formulation. Provides bus voltage magnitude + angle
+# variables, slack-bus pinning (vm + va), and active+reactive nodal balance.
+# Branch flow variables, rate limits, and AC power flow constraints are added
+# by the branch device construction path (Phase F).
+#
+# Convention: variables/constraints indexed by ACBus use bus NAME (String) — see dcp_model.jl.
+# add_variables!(VoltageAngle, ...) for ACPPowerModel is defined in dcp_model.jl
```
jd-lara, line 7: For situation like this use a network_common.jl file

src/network_models/acp_model.jl
```diff
@@ -0,0 +1,166 @@
+# Native ACP network formulation. Provides bus voltage magnitude + angle
```
jd-lara, line 1: Check that this is tested with systems that contain multiple subnetworks with distinct slack buses

src/network_models/acp_model.jl
```diff
@@ -0,0 +1,166 @@
...
+        ReferenceBusConstraint,
+        PSY.ACBus,
+        subnet_keys,
+        time_steps;
+        meta = "vm",
+    )
+
+    for k in subnet_keys
+        bus_set = subnets[k]
+        ref = _find_reference_bus(sys, bus_set)
```
jd-lara: This should error earlier in the network model subnets. This code should be already covered in several of the network models checks.

src/network_models/acp_model.jl
```diff
@@ -0,0 +1,166 @@
...
+        # Skip subnetworks without an AC reference bus (e.g. an isolated DC-side
+        # island connected through HVDC converters).
+        ref === nothing && continue
+        ref_name = ref.name
```
jd-lara: using dot access is a clear break from the Sienna.md practices. Unacceptable

src/network_models/dcp_model.jl
```diff
@@ -0,0 +1,131 @@
...
+# slack-bus pinning, and active-power nodal balance. Branch flow variables,
+# rate limits, and ohms are added by the branch device construction path
+# (see ac_transmission_models/AC_branches.jl and branch_constructor.jl).
+#
+# Convention: variables and constraints indexed by ACBus use bus NAME (String)
+# axes — matches the bridge convention and IOM's dual-extraction logic.
+# System-balance expressions (ActivePowerBalance) are indexed by bus NUMBER
+# (Int) per make_system_expressions.jl, so internal lookups translate names↔numbers.
+
+function _bus_name_number_pairs(
```
jd-lara, line 11: This information is also available from the NetworkModel

src/ac_transmission_models/AC_branches.jl
```diff
@@ -1253,3 +1269,997 @@ function add_to_objective_function!(
...
+    branch_admittance(branch) -> NamedTuple
+
+Returns the π-equivalent admittance parameters of an AC branch in per-unit:
+- `g, b`: series conductance and susceptance computed from `(r, x)`
+- `g_fr, b_fr`: from-side shunt
+- `g_to, b_to`: to-side shunt
+- `tap`: voltage-ratio magnitude (1.0 if not a tap-changing transformer)
+- `shift`: nominal phase-shift angle in radians (0.0 if not a PST; the value of α for fixed PSTs)
+"""
+function branch_admittance end
```
luke-kiernan: Does `branch_admittance` really belong here in POM? I'd sooner put it in PNM or perhaps even in PSY for the raw components.

src/ac_transmission_models/AC_branches.jl
```diff
@@ -1253,3 +1269,997 @@ function add_to_objective_function!(
...
+
+################################## Native ACP apparent-power rate constraints ###############
+
+"""
+Add from-to apparent-power rate limit constraint for ACBranch under the native ACPPowerModel.
+
+Constrains pft^2 + qft^2 ≤ rating^2.  Does not depend on PTDF / network-reduction
+infrastructure; iterates directly over devices.
+"""
+function add_constraints!(
```
luke-kiernan: These next 2 functions only differ by `FromTo` versus `ToFrom`. Combine: either multiple dispatch pattern or just separate out the 3 `get_variable` calls and delegate the last dozen lines to a helper.

src/ac_transmission_models/AC_branches.jl
```diff
@@ -1253,3 +1269,997 @@ function add_to_objective_function!(
...
+
+################################## Native DCP branch constraints ############################
+
+"""
+Add branch flow rate (rating) constraints for ACBranch under the native DCPPowerModel.
+
+This is a simple lb/ub pair on the FlowActivePowerVariable that does not depend on the
+PTDF / network-reduction infrastructure used by the AbstractActivePowerModel dispatch.
+"""
+function add_constraints!(
```
luke-kiernan, line 1653: There's nothing novel about the types or math here: imo this ought to be calling some IOM function. In the no-slacks case, `add_range_constraints!` works; we might need to add a variant of `add_range_constraints!` that allows for slacks.

src/ac_transmission_models/AC_branches.jl
```diff
@@ -1253,3 +1269,997 @@ function add_to_objective_function!(
...
+end
+
+"""
+Add branch angle-difference limit constraints for ACBranch under the native ACPPowerModel.
+
+Only branches for which `PSY.get_angle_limits` is defined and that carry non-trivial limits
+(i.e. not the ±π defaults) receive a constraint.  Branches where the method is not defined
+are silently skipped.
+"""
+function add_constraints!(
```
luke-kiernan: Looks identical or nearly identical to the `DCPPowerModel` implementation. Combine.

src/ac_transmission_models/AC_branches.jl
```diff
@@ -1253,3 +1269,997 @@ function add_to_objective_function!(
...
+#
+# Indexing the flow containers by these unique strings keeps the variable
+# storage 2D (name × time) without inventing a new container shape.
+################################################################################
+
+"""
+Returns a tuple of 3 NamedTuples, one per winding of a Transformer3W:
+  (suffix, arc, r, x, rating, tap, base_power)
+"""
+function _three_winding_arcs(t::PSY.Transformer3W)
```
luke-kiernan: `_three_winding_arcs` and `_winding_admittance`  seem like they belong in PNM or PSY.

src/network_models/dcp_model.jl
```diff
@@ -0,0 +1,131 @@
...
+        bus_by_no = Dict{Int, PSY.ACBus}(
+            PSY.get_number(b) => b for b in PSY.get_components(PSY.ACBus, sys)
+        )
+        return Tuple{String, Int}[(PSY.get_name(bus_by_no[n]), n) for n in bus_numbers]
+    end
+end
+
+# Shared between DCPPowerModel and ACPPowerModel — both put VoltageAngle on every
+# bus axis with no bounds. Slack-bus pinning is applied by ReferenceBusConstraint.
+function add_variables!(
```
luke-kiernan, line 53: Not needed: looks equivalent to the generic `add_variables!` implementation in IOM.

edit: might need to define `get_variable_binary` to be `false` for that implementation to work.

src/network_models/dcp_model.jl
```diff
@@ -0,0 +1,131 @@
...
+function _find_reference_bus_name(sys::PSY.System, bus_set)::Union{Nothing, String}
+    for b in PSY.get_components(PSY.ACBus, sys)
+        if PSY.get_number(b) in bus_set && PSY.get_bustype(b) == PSY.ACBusTypes.REF
+            return PSY.get_name(b)
+        end
+    end
+    return nothing
+end
+
+function add_constraints!(
```
luke-kiernan, line 123: Same as above: check IOM for relevant helper functions

src/ac_transmission_models/AC_branches.jl
```diff
@@ -1253,3 +1269,997 @@ function add_to_objective_function!(
...
+        b_to = adm.b_to
+        tm = adm.tap
+        nominal_shift = adm.shift
+        from_bus_obj = PSY.get_from(PSY.get_arc(d))
+        to_bus_obj = PSY.get_to(PSY.get_arc(d))
+        from_bus = PSY.get_name(from_bus_obj)
+        to_bus = PSY.get_name(to_bus_obj)
+        # Pre-compute constant coefficients from tap + shift.
+        # Convention (same as PowerModels.jl ACP):
+        #   tr = tm * cos(shift),  ti = tm * sin(shift)
```
luke-kiernan, line 1878: Yikes this is complicated. Would be good to get some tests here. Is there a simple way to tell if it's correct via complex number math or numerical approximations?

luke-kiernan (commented):
Big picture things:
1. In my opinion, `JuMP` calls only belong in POM if they're device-specific or otherwise bespoke. If you're just assigning upper/lower bounds with some slacks, I'd prefer to have the math live back in IOM.
2. In `construct_device!` calls, we almost always write
```julia
add_variables!(container, FlowActivePowerFromToVariable, devices, StaticBranchBounds)
add_variables!(container, FlowActivePowerToFromVariable, devices, StaticBranchBounds)
add_variables!(container, FlowReactivePowerFromToVariable, devices, StaticBranchBounds)
add_variables!(container, FlowReactivePowerToFromVariable, devices, StaticBranchBounds)
```
instead of
```julia
for flow_var in (FlowActivePowerFromToVariable, ...) # list 4 variable types
    add_variables!(container, flow_var, devices, StaticBranchBounds)
end
```
Is there a good reason for this pattern? I find the 2nd much easier to parse at a glance

jd-lara:
@luke-kiernan after a second look this PR needs more work since it isn't leveraging correctly the mappings in NetworkModels. I need to direct Claude better

PR #112, opened 2026-05-06
Cost expression refactor, parameter broadcasting, and PF headroom slack porting

### LINKED ISSUE #106: Move over PSI PR #1609
Move over the performance improvements from PSI PR [#1609](https://github.com/Sienna-Platform/PowerSimulations.jl/pull/1609).

### LINKED ISSUE #109: Implement power flow in-the-loop
In the initial IOM-POM refactor, I skipped over the power flow in the loop code and put a "yet to be implemented" stub. Fix that: move over the relevant code from PSI to here in POM and write some tests.

acostarelli:
https://github.com/Sienna-Platform/PowerSimulations.jl/pull/1564
https://github.com/Sienna-Platform/PowerSimulations.jl/pull/1609
https://github.com/Sienna-Platform/PowerSimulations.jl/pull/1566

Requires IOM on https://github.com/Sienna-Platform/InfrastructureOptimizationModels.jl/pull/97


ext/PowerFlowsExt/pf_solve_and_aux.jl
```diff
@@ -0,0 +1,155 @@
+# Power flow in-the-loop: solve dispatcher and auxiliary variable readback.
+# Ported from PowerSimulations.jl/src/network_models/power_flow_evaluation.jl
+# (lines 753-874). Defines latest_solved_power_flow_evaluation_data, solve_powerflow!
+# (renamed from PSI's solve_power_flow!), calculate_aux_variable_value! overloads
+# for PowerFlowAuxVariableType, and _get_pf_result helpers.
+
+"Fetch the most recently solved `PowerFlowEvaluationData`"
+function latest_solved_power_flow_evaluation_data(container::OptimizationContainer)
+    datas = get_power_flow_evaluation_data(container)
+    return datas[findlast(x -> x.is_solved, datas)]
```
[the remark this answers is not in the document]
luke-kiernan: Agreed: we should address this edge case and add a test case that exercises the code path.

ext/PowerFlowsExt/pf_solve_and_aux.jl
```diff
@@ -0,0 +1,155 @@
...
+end
+
+function solve_powerflow!(
```
luke-kiernan: Nitpick: I'd prefer to keep the names consistent and call this `solve_power_flow!`.

ext/PowerFlowsExt/pf_solve_and_aux.jl
```diff
@@ -0,0 +1,155 @@
...
+
+function IOM.calculate_aux_variable_value!(
+    container::OptimizationContainer,
+    key::AuxVarKey{<:POM.PowerFlowAuxVariableType, <:PSY.Component},
+    system::PSY.System,
+)
+    # Skip the aux vars that the current power flow isn't meant to update
+    pf_e_data = latest_solved_power_flow_evaluation_data(container)
+    pf_data = get_power_flow_data(pf_e_data)
+    (key in branch_aux_vars(pf_data) || key in bus_aux_vars(pf_data)) && return
```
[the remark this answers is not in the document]
luke-kiernan: Yeah actual issue. IOM/POM often pass types types, where PSI passes instances, so by copy-pasting this over, you've mixed the two. Also please write a test case for this path.

src/common_models/add_expressions.jl
```diff
@@ -61,14 +75,12 @@ function add_expressions!(
...
-        op_cost = PSY.get_operation_cost(d)
-        fuel_curve = _get_variable_if_exists(op_cost)
-        if fuel_curve isa PSY.FuelCurve
-            push!(names, PSY.get_name(d))
-            if !found_quad_fuel_functions
-                found_quad_fuel_functions =
-                    PSY.get_value_curve(fuel_curve) isa PSY.QuadraticCurve
-            end
+        fuel_curve = _get_variable_if_exists(PSY.get_operation_cost(d))
+        _is_fuel_curve(fuel_curve) || continue
```
luke-kiernan, line 79: Why is this 2nd check necessary?

Also, I find the name `_get_variable_if_exists` misleading. (It's a cost function thing, not an optimization key thing, yet the name starts with `get_variable`.) Consider re-naming

src/common_models/add_parameters.jl
```diff
@@ -200,21 +200,26 @@ function _add_time_series_parameters!(
 
     jump_model = get_jump_model(container)
     param_instance = T()
-    for (ts_uuid, raw_ts_vals) in initial_values
+    parent_param = IOM.get_parameter_array_data(param_container)
+    # `collect(keys(initial_values))` was passed to `add_param_container!` above as the
+    # parameter axis, so iterating `initial_values` in order matches the container's
+    # first axis row-by-row.
+    for (i, (ts_uuid, raw_ts_vals)) in enumerate(initial_values)
         ts_vals = _unwrap_for_param.(Ref(param_instance), raw_ts_vals, Ref(additional_axes))
```
luke-kiernan: Nitpick: the `Ref(foo)`'s here are for broadcasting purposes, "treat this as an array of length one." There's reasons to prefer singleton tuple for such purposes, `(foo,)`: can be stack-allocated.

src/common_models/add_parameters.jl
```diff
@@ -200,21 +200,26 @@ function _add_time_series_parameters!(
...
+    parent_mult = IOM.get_multiplier_array_data(param_container)
+    # `devices_with_time_series` was used to build `device_names`, which is the
+    # multiplier array's first axis, so enumeration index `i` lines up.
+    for (i, device) in enumerate(devices_with_time_series)
         multiplier = get_multiplier_value(T, device, W)
         device_name = PSY.get_name(device)
-        for step in time_steps
-            IOM.set_multiplier!(param_container, multiplier, device_name, step)
-        end
+        IOM._set_multiplier_at!(parent_mult, Float64(multiplier), i)
```
luke-kiernan: Nitpick: `Float64` conversion here should be avoidable

src/common_models/add_parameters.jl
luke-kiernan:
```suggestion
    devices::Union{Vector{D}, IS.FlattenIteratorWrapper{D}}
```
pre-existing, but I'm puzzled by the fact that this is untyped, even though we have `for device::D in devices`

luke-kiernan, line 263: Ick we should open an issue for this

luke-kiernan: Same here, curiously untyped argument
```suggestion
    devices::Union{Vector{D}, IS.FlattenIteratorWrapper{D}}
```

src/common_models/add_to_expression.jl
```diff
@@ -2627,62 +2696,85 @@ function add_to_expression!(
...
+            JuMP.add_to_expression!(
+                expression[name, t], P_min * prop_term_per_unit * dt, bin)
+        elseif sos_status == SOSStatusVariable.VARIABLE
+            var = get_default_on_variable(d)
+            bin = get_variable(container, var, V)[name, t]
+            JuMP.add_to_expression!(
+                expression[name, t], P_min * prop_term_per_unit * dt, bin)
+        else
+            @assert false
         end
```
luke-kiernan: Call `_determine_bin_lhs` in IOM instead: up to hard-coded `var = OnVariable` there vs `var = get_default_on_variable(d)` here, this is
```julia
bin = _determine_bin_lhs(container, sos_status, V, name, t)
JuMP.add_to_expression!(expression[name, t], P_min * prop_term_per_unit * dt, bin)
```

src/common_models/add_to_expression.jl
```diff
@@ -2609,6 +2609,75 @@ function add_cost_to_expression!(
     return
 end
 
+# Per-device fuel consumption term builders, dispatched on the value-curve type so the
+# decision of how to translate the curve into JuMP terms is a method-resolution problem
+# rather than a runtime branch.
+
+# LinearCurve fuel: linear in the dispatch variable.
+function _add_fuel_consumption_term!(
```
luke-kiernan, line 2658: Identical to IOM's `add_proportional_cost_invariant!`, except we skip the final step where we add to the objective function and to the `ProductionCostExpression`. Refactor?

luke-kiernan (commented):
Big picture things:

1. Just generally: before adding code to POM, ask yourself "is there device specific handling here?" If the answer is "no," it probably belongs better in IOM. And there may be a helper for it there already.
2. All the expression stuff is currently in POM, not in IOM, whether device-specific or generic. That's purely due to copy-pasting from PSI, not due to intentional organization choices. We should probably change that at some point and move the generic implementations into IOM.
3. Test coverage. I ran the tests locally and they pass, but with the level of complexity here, I wouldn't be surprised if some amount of the code isn't actually getting run. Lots of different multiple dispatch cases: e.g. `_set_parameter_at` with HDF5 storage.

src/common_models/objective_function.jl
```diff
@@ -15,3 +15,90 @@ function add_variable_cost_to_objective!(
...
+    )
+    resolution = get_resolution(container)
+    dt = Dates.value(resolution) / MILLISECONDS_IN_HOUR
+    name = PSY.get_name(component)
+    dispatch_vars = get_variable(container, T, C)
+
+    for t in get_time_steps(container)
+        offer_max = _renewable_offer_max(container, component, name, t)
+        dispatch = dispatch_vars[name, t]
+        curtailment_cost = proportional_term_per_unit * dt * (offer_max - dispatch)
```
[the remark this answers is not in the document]
acostarelli: @m-bossart This was also missing from your PR, is this intentional?

ext/PowerFlowsExt/pf_solve_and_aux.jl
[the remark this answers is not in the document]
acostarelli: Would this ever happen? Optimization container solves PF before calculating aux variables.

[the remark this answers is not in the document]
luke-kiernan: > Would this ever happen? Optimization container solves PF before calculating aux variables.

Yes! AC power flows can fail to converge. And if no time step converges, this function errors.

The optimization problems we're solving in IOM/POM often use a simplified transmission model and ignore reactive power. An AC power flow models those things in full detail. 

For a concrete example, make a network with super high resistance lines and run a simulation with a `CopperPlate` network model. Generation is adequate if you pretend everything is lossless (optimization problem solves), but not enough power will reach the loads due to transmission issues (AC power flow fails).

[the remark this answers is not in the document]
luke-kiernan: Oh wait...
```julia

function solve_power_flow!(...)
    if (...)
        PFS.solve_power_flow!(pf_data)
        ...
    end
    pf_e_data.is_solved = true
    return
end
```
So with the code as-is, yeah, this is unreachable. The correct thing would be to set `pf_e_data.is_solved = PFS.solve_power_flow!(pf_data)`. But design choices around how to handle failure were wide open at the time (and still are kinda up-in-the-air), so someone hard-coded "let's just assume the power flow converged."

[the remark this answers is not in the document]
luke-kiernan: I guess I'd just add a `# FIXME` comment in both spots (hard-coded `is_solved = true` and `findlast` call) and leave things as-is, since "how to handle failure" is something that merits further discussion

luke-kiernan (approved):
Looks good. I'm noticing the following sequence of calls in a couple different places
```julia
power_units = PSY.get_power_units(var_cost)
proportional_term = PSY.get_proportional_term(value_curve)
prop_term_per_unit = get_proportional_cost_per_system_unit(
    proportional_term, power_units, base_power, device_base_power)
rate = prop_term_per_unit * dt # sometimes also * multipler.
```
 but that's something we can address later

docs/Project.toml
```diff
@@ -7,6 +7,12 @@ Literate = "98b081ad-f1c9-55d3-8b20-4c87d4299306"
 PowerOperationsModels = "bed98974-b02a-5e2f-9ee0-a103f5c450dd"
 PrettyTables = "08abe8d2-0d0c-5749-adfa-8a2ac140af0d"
 
+[sources]
+PowerOperationsModels = {path = ".."}
```
luke-kiernan: Do we want this POM pin or is this a temporary testing thing? Not sure--would have to check other projects for how they handle the docs environment.

ext/PowerFlowsExt/pf_input_mapping.jl
```diff
@@ -316,16 +316,21 @@ _with_time_steps(pf::PFS.PSSEExportPowerFlow, ::Int) = pf
 
 function POM.add_power_flow_data!(
     container::OptimizationContainer,
-    evaluators::Vector{<:PFS.PowerFlowEvaluationModel},
+    network_model::IOM.NetworkModel,
     sys::PSY.System,
 )
-    container.power_flow_evaluation_data = Vector{PowerFlowEvaluationData}()
-    sizehint!(container.power_flow_evaluation_data, length(evaluators))
+    evaluations = get_evaluations(network_model)
```
luke-kiernan, line 322: >     # Share the same EvaluationContainer instance between NetworkModel and
>     # OptimizationContainer so IOM.calculate_aux_variables! sees the same
>     # runtime state we register here.
Is this a bugfix or consequence of how we're restructuring things? Are these also shared in PSI?

src/common_models/objective_function.jl
```diff
@@ -15,3 +15,92 @@ function add_variable_cost_to_objective!(
...
+function _renewable_offer_max(
+    container::OptimizationContainer,
+    component::C,
+    name::String,
+    t::Int,
+) where {C <: PSY.RenewableGen}
+    has_container_key(container, ActivePowerTimeSeriesParameter, C) ||
+        return PSY.get_max_active_power(component)
+    param_container = get_parameter(container, ActivePowerTimeSeriesParameter, C)
+    multiplier = get_multiplier_array(param_container)
```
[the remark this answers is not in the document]
luke-kiernan: Good catch, yeah check this

src/core/interfaces.jl
```diff
@@ -146,31 +146,42 @@ end
...
     )
     return
 end
 
+"""
+Build an `EvaluationContainer` holding a single evaluator. Convenience for the
+common single-evaluator case at call sites such as
+`NetworkModel(...; evaluations = power_flow_evaluations(ACPowerFlow()))`.
+"""
+function power_flow_evaluations(ev)
```
luke-kiernan: > `typeof(ev)`

Switch to a type parameter, `ev::T`.

src/core/problem_template.jl
```diff
@@ -361,7 +364,7 @@ function _populate_aggregated_service_model!(
     return
 end
 
-function finalize_template!(template::OperationsProblemTemplate, sys::PSY.System)
+function finalize_template!(template::PowerOperationsProblemTemplate, sys::PSY.System)
```
luke-kiernan, line 374: clarifying question: do we have a separate `OperationsProblemTemplate` that prompted the rename?

src/core/problem_types.jl
```diff
@@ -0,0 +1,53 @@
...
+Supertype for rolling-horizon (emulation) operation problems solved by `EmulationModel`.
+"""
+abstract type AbstractPowerEmulationProblem <: AbstractPowerOperationProblem end
+
+"""
+Decision problems whose build/validate behavior is fully driven by the
+`ProblemTemplate` (the common case). `DecisionModel{<:DefaultPowerDecisionProblem}`
+gets the generic template-driven `build_model!`/`validate_template` methods.
+"""
+abstract type DefaultPowerDecisionProblem <: AbstractPowerDecisionProblem end
```
luke-kiernan: nitpick on naming clarity: default sounds more specific than generic. So I'd be inclined to switch the order, have `Generic` as the supertype and `Default` as the instance.

src/initial_conditions/initialization.jl
```diff
@@ -17,16 +20,16 @@ function get_initial_conditions_template(model::OperationModel, number_of_steps:
     network_model.network_reduction =
         deepcopy(get_network_reduction(get_network_model(model.template)))
     network_model.subnetworks = get_subnetworks(get_network_model(model.template))
-    # Initialization does not support PowerFlow evaluation - use empty vector
-    network_model.power_flow_evaluation = AbstractPowerFlowEvaluationModel[]
+    # Initialization does not run external evaluators
```
luke-kiernan: >  `# Initialization does not run external evaluators`

clarifying question: are there any internal evaluators? I don't think so, which makes the comment a bit deceptive

src/operation/decision_model.jl
```diff
@@ -1,4 +1,28 @@
...
+function DecisionModel(
+    template::IOM.AbstractProblemTemplate,
+    sys::IS.InfrastructureSystemsContainer,
+    jump_model::Union{Nothing, JuMP.Model} = nothing;
+    kwargs...,
+)
+    return DecisionModel{GenericPowerDecisionProblem}(template, sys, jump_model; kwargs...)
+end
+
+# Template-driven problems require a template; the bare-system constructor is an error.
```
luke-kiernan: Confusing comment imo. Just say "default decision problems require a `AbstractProblemTemplate` subtype."

(Are there other problems that don't require a template? If so, which?)

src/operation/emulation_model.jl
```diff
@@ -1,10 +1,67 @@
...
+        template,
+        sys,
+        jump_model;
+        kwargs...,
+    )
+end
+
+# POM-side implementation of IOM's `validate_time_series!` extension point for
+# emulation problems. Emulation models solve a single step, so horizon == resolution.
+function validate_time_series!(model::EmulationModel{<:DefaultPowerEmulationProblem})
```
luke-kiernan: could combine with decision problem to cut down on code repetition

src/static_injector_models/electric_loads.jl
```diff
@@ -236,7 +236,11 @@ function add_constraints!(
     jump_model = get_jump_model(container)
     for t in time_steps, d in devices
         name = PSY.get_name(d)
-        pf = sin(atan((PSY.get_max_reactive_power(d) / PSY.get_max_active_power(d))))
+        pf = sin(
+            atan((
+                PSY.get_max_reactive_power(d, PSY.SU) / PSY.get_max_active_power(d, PSY.SU)
+            )),
```
luke-kiernan: Has closed form via Pythagorean identities: if `r` is for reactive and `a` for active, I'm getting `r/sqrt(r^2+a^2)`

luke-kiernan (changes requested):
Mostly small things. The 2 that require attention/discussion: CoPilot's comment on `_renewable_offer_max` and sharing the `EvaluationContainer`.

docs/Project.toml
acostarelli: Testing thing will remove

src/core/problem_template.jl
acostarelli: No just making clear that "Operations" refers to "Power Operations" specifically.

src/core/problem_types.jl
acostarelli: Agreed the naming is confusing I'll see how that looks

src/operation/decision_model.jl
acostarelli: Yeah this comment confused me too

ext/PowerFlowsExt/pf_solve_and_aux.jl
```diff
@@ -125,15 +125,16 @@ function IOM.calculate_aux_variable_value!(
...
         for br in parallel_brs
-            sample_line = first(parallel_brs)
-            impedance = PSY.get_r(sample_line) + im * PSY.get_x(sample_line)
-            first_name = PSY.get_name(sample_line)
             if br isa U
                 name = PSY.get_name(br)
                 IS.@assert_op T <: POM.BranchFlowAuxVariableType ||
                               (T == POM.PowerFlowBranchActivePowerLoss)
-                if !isapprox(PSY.get_r(br) + im * PSY.get_x(br), impedance)
+                if !isapprox(PSY.get_r(br, PSY.SU) + im * PSY.get_x(br, PSY.SU), impedance)
```
luke-kiernan: Still needed? With updating PSI to be compatible with network reductions, I think it might be fine: see `compute_parallel_multiplier` in PNM.

src/twoterminal_hvdc_models/AC_branches.jl
```diff
@@ -255,7 +255,7 @@ end
 
 ################################## Rate Limits constraint_infos ############################
 
-function get_rating(double_circuit::PNM.BranchesParallel)
+function get_rating(double_circuit::PNM.AbstractBranchesParallel)
```
luke-kiernan: nitpick: looks near-identical to `PNM.get_sum_of_max_rating`

test/test_market_bid_cost.jl
```diff
@@ -13,10 +13,21 @@ numbers put on a cost curve reach the container's objective coefficients as expe
 const _LOAD_NAME = "load1"
 const _THERMAL_NAME = "thermal1"
 
-# A static MBC with a decremental offer curve only (incremental stays at the default
-# ZERO_OFFER_CURVE, which the load-side supply check treats as "absent").
+# A zero offer curve in SYSTEM_BASE. Used as the "absent" side when only one of the
+# incremental/decremental curves is meaningful: PSY requires both offer curves to share
+# a unit system, and the default `ZERO_OFFER_CURVE` is NATURAL_UNITS, so we can't leave
+# the other side defaulted when the meaningful side is SYSTEM_BASE. A zero value curve is
```
luke-kiernan, line 19: Should we go change the default in PSY? Could open an issue there to discuss--if most curves are in SU, then imo we should.

Project.toml
```diff
@@ -22,12 +22,13 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
...
 
 [weakdeps]
-PowerFlows = "94fada2c-0ca5-4b90-a1fb-4bc5b59ccfc7"
+PowerFlows = "94fada2c-fd9a-4e89-8d82-81405f5cb4f6"
 
 [sources]
 InfrastructureSystems = {rev = "IS4", url = "https://github.com/Sienna-Platform/InfrastructureSystems.jl"}
 PowerSystems = {rev = "psy6", url = "https://github.com/Sienna-Platform/PowerSystems.jl"}
-InfrastructureOptimizationModels = {rev = "main", url = "https://github.com/Sienna-Platform/InfrastructureOptimizationModels.jl"}
+InfrastructureOptimizationModels = {path = "InfrastructureOptimizationModels.jl"}
```
luke-kiernan: I'd be surprised if the CI passes with this environment: you probably need IOM, PF, and PNM pinned to the correct branches (and not to local paths).

src/core/problem_template.jl
```diff
@@ -40,29 +40,35 @@ function Base.isempty(template::OperationsProblemTemplate)
...
+function get_component_types(template::PowerOperationsProblemTemplate)::Vector{Type}
     return vcat(
         get_component_type.(values(get_device_models(template))),
         get_component_type.(values(get_branch_models(template))),
         get_component_type.(values(get_service_models(template))),
     )
 end
 
-function get_model(template::OperationsProblemTemplate, ::Type{T}) where {T <: PSY.Device}
+function get_model(
```
luke-kiernan, line 68: (Pre-existing issue.) Where do we use `get_model`? Between here and `main` branch of IOM, the only call sites I can find are in `test/test_utils/model_checks.jl`. Perhaps we need it for the eventual PSI that sits on top of POM and IOM?

src/common_models/add_expressions.jl
```diff
@@ -11,8 +11,22 @@ function _ref_index(::NetworkModel{AreaPTDFPowerModel}, device_bus::PSY.ACBus)
...
+
+# Predicates for fuel-curve detection. Dispatch over the value returned by
+# `_get_cost_if_exists` so callers can avoid `isa` checks.
+_is_fuel_curve(::Nothing) = false
+_is_fuel_curve(::PSY.CostCurve) = false
+_is_fuel_curve(::PSY.FuelCurve) = true
+
+# Predicates for value-curve shape. Used to decide whether a FuelConsumptionExpression
+# container needs JuMP.QuadExpr storage (vs the cheaper GAE).
+_value_curve_is_quadratic(::PSY.LinearCurve) = false
```
luke-kiernan, line 25: Could just write
```julia
_value_curve_is_quadratic(::PSY.InputOutputCurve{QuadraticFunctionData}) = true
_value_curve_is_quadratic(::PSY.InputOutputCurve{<:FunctionData}) = false
```
`IS.FunctionData` is mostly an internal type, though, so this is perhaps a little bit less-than-kosher.

src/ac_transmission_models/AC_branches.jl
```diff
@@ -260,14 +260,14 @@ end
...
 
-function get_rating(double_circuit::PNM.BranchesParallel)
-    return sum([PSY.get_rating(circuit) for circuit in double_circuit])
+function get_rating(double_circuit::PNM.AbstractBranchesParallel)
+    return sum([PSY.get_rating(circuit, PSY.SU) for circuit in double_circuit])
 end
 function get_rating(series_chain::PNM.BranchesSeries)
     return minimum([get_rating(segment) for segment in series_chain])
 end
 function get_rating(device::T) where {T <: PSY.ACTransmission}
```
luke-kiernan: I like the forced-function of having phased-out functions (like single-argument `get_rating`) error. I'd vote for re-naming this function to `_get_rating`: internal only.

Consider the following:
```julia
using PowerSimulations
from PowerSimulations import get_rating
using PowerSystems
from PowerSystems import get_rating
# set g equal to some generator from sys.
set_units_base_system!(sys, "DEVICE_BASE")
# pre-unitful PSY6: returns rating in device base, via PSY's single arg get_rating
# post unitful PSY6: returns rating in system base, via PSI's single arg get_rating.
r = get_rating(g)
```
Now, we're not exporting PSI's `get_rating`, nor importing PSY's `get_rating`. So you'd really have to try to trip upon this. But still, my vote would be to rename, to avoid confusion and bugs.

luke-kiernan: I asked Claude to search for similar name collisions: `get_base_power` is close, but PSI's 1-arg version only accepts `OptimizationContainer` and `ModelStoreParams` types.

luke-kiernan (commented):
Second pass: the `get_rating` name collision here is the big one

jd-lara (approved):

PR #119, opened 2026-05-12
Add HVDCTwoTerminalVSC two-terminal HVDC formulation

acostarelli:


src/ac_transmission_models/branch_constructor.jl
```diff
@@ -1670,3 +1670,180 @@ function construct_device!(
...
     return
 end
+
+############################################################################
+####################### Two-Terminal VSC HVDC Construct ####################
+############################################################################
+
+# Quadratic / bilinear approximation traits — same scheme used by the MT
+# converter formulations.
+_quad_config(::Type{HVDCTwoTerminalVSC}) = IOM.NoQuadApproxConfig()
```
acostarelli: Are these getting reused by the QuadraticLossConverter?

src/ac_transmission_models/branch_constructor.jl
```diff
@@ -1670,3 +1670,180 @@ function construct_device!(
...
+    IOM.Bin2Config(IOM.SolverSOS2QuadConfig(DEFAULT_INTERPOLATION_LENGTH))
+
+# IOM's quadratic/bilinear approximation helpers take per-device bounds as
+# `Vector{IOM.MinMax}` (one (min, max) pair per device, same order as `devices`).
+function _vsc_v_from_bounds(devices)
+    n = length(devices)
+    bounds = Vector{IOM.MinMax}(undef, n)
+    for (k, d) in enumerate(devices)
+        lims = PSY.get_voltage_limits_from(d)
+        bounds[k] = IOM.MinMax((min = lims.min, max = lims.max))
```
acostarelli: We don't need to reference IOM.MinMax explicitly.

src/common_models/quadratic_converter_loss.jl
```diff
@@ -0,0 +1,124 @@
...
+#########################################
+######## Loss expression builder ########
+#########################################
+
+# Returns the JuMP expression  a*i_sq + b*(i_pos + i_neg) + c
+# for a single (device, time). The b*(i_pos+i_neg) term is included only when
+# the formulation has opted into the linear-loss path AND b is nonzero for
+# this specific device.
+function _quadratic_converter_loss_expr(
+    a, b, c, i_sq_t, i_pos_t, i_neg_t; use_linear_loss::Bool,
```
acostarelli: Can we add types to these parameters?

src/common_models/quadratic_converter_loss.jl
```diff
@@ -0,0 +1,124 @@
...
+# HVDCTwoTerminalVSC formulations.
+
+#########################################
+######## Loss-curve introspection #######
+#########################################
+
+_get_quadratic_term(loss_fn::PSY.QuadraticCurve) = PSY.get_quadratic_term(loss_fn)
+_get_quadratic_term(loss_fn) = 0.0
+
+# Whether the formulation wants the b*|I| linear-loss term (which requires
```
acostarelli: We should be clear about defaults. If I understand correctly, the only reason why we wouldn't use linear loss is if we're using Ipopt on an NLP, which is why we expose `use_linear_loss` as an attribute for the NLP because the user might use Gurobi. Otherwise, if we're building a MILP, then we'll just default to saying yes we will use linear loss.

src/common_models/quadratic_converter_loss.jl
```diff
@@ -0,0 +1,124 @@
...
+# Dispatched on device type because different PSY devices store loss curves on
+# different fields.
+_has_linear_loss(d::PSY.InterconnectingConverter) =
+    !iszero(PSY.get_proportional_term(PSY.get_loss_function(d)))
+_has_linear_loss(d::PSY.TwoTerminalVSCLine) =
+    !iszero(PSY.get_proportional_term(PSY.get_converter_loss_from(d))) ||
+    !iszero(PSY.get_proportional_term(PSY.get_converter_loss_to(d)))
+
+function _devices_with_linear_loss(devices)
+    return [d for d in devices if _has_linear_loss(d)]
```
acostarelli: Is there a built-in filter function that might be more performant?

src/common_models/quadratic_converter_loss.jl
```diff
@@ -0,0 +1,124 @@
...
+#########################################
+
+# Returns the JuMP expression  a*i_sq + b*(i_pos + i_neg) + c
+# for a single (device, time). The b*(i_pos+i_neg) term is included only when
+# the formulation has opted into the linear-loss path AND b is nonzero for
+# this specific device.
+function _quadratic_converter_loss_expr(
+    a, b, c, i_sq_t, i_pos_t, i_neg_t; use_linear_loss::Bool,
+)
+    loss = a * i_sq_t + c
```
acostarelli: Do we also want to check if `a` is 0? So we're not putting 0s into the expression -- which is not great numerically, right? Feel free to push back on this.

src/common_models/quadratic_converter_loss.jl
```diff
@@ -0,0 +1,124 @@
...
+end
+
+#########################################
+####### Abs-value decomposition #########
+#########################################
+
+# Adds the three variables (PositiveCurrent, NegativeCurrent, CurrentDirection)
+# that decompose a signed current variable into  i = i^+ - i^-  with a binary
+# direction indicator. Called from the ArgumentConstructStage.
+function _add_abs_value_decomposition_variables!(
```
acostarelli, line 57: Why is this a separate routine than adding the constraints?

src/common_models/quadratic_converter_loss.jl
```diff
@@ -0,0 +1,124 @@
...
+) where {D <: PSY.Device, F}
+    time_steps = get_time_steps(container)
+    names = [PSY.get_name(d) for d in devices]
+    jump_model = get_jump_model(container)
+    i_var = get_variable(container, parent_var_type, D)
+    i_pos_var = get_variable(container, PositiveCurrent, D)
+    i_neg_var = get_variable(container, NegativeCurrent, D)
+    i_dir_var = get_variable(container, CurrentDirection, D)
+
+    abs_val_const = add_constraints_container!(
```
acostarelli, line 100: Is it better to have 3 constraints with different meta strings here, or 1 container with a 3-length axis?

src/network_models/pm_translator.jl
```diff
@@ -720,10 +741,7 @@ function get_branches_to_pm(
     for (d, device_model) in branch_template
         comp_type = get_component_type(device_model)
         !(comp_type <: T) && continue
```
acostarelli, line 743: Not part of this PR, but is there a clean way to replace this type check with multiple dispatch?

src/twoterminal_hvdc_models/TwoTerminalDC_branches.jl
```diff
@@ -1433,3 +1433,362 @@ function add_constraints!(
...
+                jump_model,
+                p_tf[name, t] == -vi_expr_to[name, t] + loss_tf,
+            )
+        end
+    end
+    return
+end
+
+# PQ capability:  p_k^2 + q_k^2 <= S_k^2 (NLP) or octagonal polygon (Bin2).
+function add_constraints!(
```
acostarelli, line 1667: Can this be combined with the Bin2 dispatch below? Does something about the IOM API make this awkward that we could change?

test/test_device_hvdc.jl
```diff
@@ -208,9 +208,147 @@ end
...
+end
+
+@testset "HVDC Two-Terminal VSC (NLP) on DCP" begin
+    model = _build_vsc_model(HVDCTwoTerminalVSC, DCPPowerModel, ipopt_optimizer)
+    @test build!(model; output_dir = mktempdir(; cleanup = true)) ==
+          IOM.ModelBuildStatus.BUILT
+    @test solve!(model) == IOM.RunStatus.SUCCESSFULLY_FINALIZED
+end
+
+@testset "HVDC Two-Terminal VSC on AC (NLP)" begin
```
acostarelli, line 262: The VSC on AC (Bin2) test was omitted. Please indicate why.

Also, it's possible that we will end up using a different MILP formulation, or that that will become an attribute. So we should probably change names that have Bin2 to be something more generic. Maybe using "MIP" or "MILP" instead.

src/twoterminal_hvdc_models/TwoTerminalDC_branches.jl
```diff
@@ -1433,3 +1433,362 @@ function add_constraints!(
...
+get_variable_upper_bound(::Type{NegativeCurrent}, d::PSY.TwoTerminalVSCLine, ::Type{<:AbstractTwoTerminalVSCFormulation}) = _vsc_shared_i_max(d)
+
+#! format: on
+
+####################### VSC reactive-power optional path #####################
+# The reactive power and PQ capability machinery is added only when the
+# network actually models reactive power. On active-only networks these
+# helpers are no-ops.
+
+function _maybe_add_reactive_power_variables!(
```
acostarelli: Maybe out of scope for this PR, but I imagine there are many formulations across POM that could use this for adding their own reactive power variables.

src/twoterminal_hvdc_models/TwoTerminalDC_branches.jl
```diff
@@ -1433,3 +1433,362 @@ function add_constraints!(
...
+    container::OptimizationContainer,
+    devices,
+    model::DeviceModel{PSY.TwoTerminalVSCLine, F},
+    ::NetworkModel{<:AbstractPowerModel},
+) where {F <: AbstractTwoTerminalVSCFormulation}
+    add_variables!(container, HVDCReactivePowerFromVariable, devices, F)
+    add_variables!(container, HVDCReactivePowerToVariable, devices, F)
```
acostarelli: Why do we have specific From and To variables for reactive power? Is it possible to just model the difference, or to use one container that has another axis, and would either of those be better?

And if the From/To separation is better, would we gain anything by using parametric types?

src/ac_transmission_models/branch_constructor.jl
```diff
@@ -1810,14 +1787,15 @@ function construct_device!(
...
-    if _use_linear_loss(F, device_model)
-        ll_devices = _devices_with_linear_loss(devices)
-        if !isempty(ll_devices)
-            _add_abs_value_decomposition_constraints!(
-                container, ll_devices, device_model, network_model,
-                DCLineCurrentFlowVariable, _vsc_shared_i_max,
-            )
-        end
+    use_ll = get_attribute(device_model, "use_linear_loss")
+    if F === HVDCTwoTerminalVSC && use_ll
```
acostarelli: Don't do explicit type-checking, use a helper.

src/common_models/add_to_expression.jl
```diff
@@ -718,6 +718,78 @@ function add_to_expression!(
...
+function add_to_expression!(
+    container::OptimizationContainer,
+    ::Type{T},
+    ::Type{U},
+    devices::IS.FlattenIteratorWrapper{V},
+    ::DeviceModel{V, W},
+    network_model::NetworkModel{X},
+) where {
+    T <: ReactivePowerBalance,
+    U <: HVDCReactivePowerVariable{From},
```
acostarelli: The idea behind parameterizing this variable was that you could combine methods like these and add a helper that dispatches on From/To.

src/common_models/quadratic_converter_loss.jl
```diff
@@ -11,79 +11,53 @@
...
 )
-    loss = a * i_sq_t + c
-    if use_linear_loss && !iszero(b)
-        loss += b * (i_pos_t + i_neg_t)
-    end
-    return loss
+    quad = iszero(a) ? 0 : a * i_sq_t
+    lin = (use_linear_loss && !iszero(b)) ? b * (i_pos_t + i_neg_t) : 0
+    const_term = iszero(c) ? 0 : c
+    return quad + lin + const_term
```
acostarelli: Are these jump expressions that we're adding? Not super efficient to be doing it this way but maybe not a bottleneck.

src/mt_hvdc_models/HVDCsystems.jl
```diff
@@ -140,18 +140,16 @@ get_variable_upper_bound(::Type{NegativeCurrent}, d::PSY.InterconnectingConverte
 
 function get_default_attributes(
     ::Type{PSY.InterconnectingConverter},
-    ::Type{Bin2QuadraticLossConverter},
+    ::Type{MIPQuadraticLossConverter},
 )
-    return Dict{String, Any}()
+    return Dict{String, Any}("use_linear_loss" => true)
```
acostarelli, line 148: Add comments here explaining why the defaults are what they are

src/mt_hvdc_models/hvdcsystems_constructor.jl
```diff
@@ -151,17 +141,16 @@ function construct_device!(
...
-                container,
-                CurrentAbsoluteValueConstraint,
-                ll_devices,
-                model,
-                network_model,
-            )
-        end
+
+    use_ll = get_attribute(model, "use_linear_loss")
+    if T === QuadraticLossConverter && use_ll
```
acostarelli: Again, don't check types explicitly. Use multiple dispatch.

src/twoterminal_hvdc_models/TwoTerminalDC_branches.jl
```diff
@@ -1433,3 +1433,387 @@ function add_constraints!(
...
+            cons_t[name, t] = JuMP.@constraint(
+                jump_model,
+                p_tf[name, t]^2 + q_t[name, t]^2 <= s_t^2,
+            )
+        end
+    end
+    return
+end
+
+function add_constraints!(
```
acostarelli, line 1721: I'm not really sure what's going on here. Are you building some envelope? Why are we not computing P^2, Q^2 and S^2 using the approximation code we have in IOM? This would also allow combining the these two add_constraints! methods.

acostarelli (commented):
Review 2

src/core/variables.jl
```diff
@@ -384,6 +387,34 @@ Docs abbreviation: ``z``
...
+struct HVDCFromDCVoltage <: VariableType end
+
+"""
+DC-side voltage at the to-terminal of a two-terminal HVDC link.
+Used by `HVDCTwoTerminalVSC` formulations.
+Docs abbreviation: ``v_t^{dc}``
+"""
+struct HVDCToDCVoltage <: VariableType end
+
+abstract type From end
```
rodrigomha: This looks wierd. Probably an hallucination

src/core/constraints.jl
```diff
@@ -419,6 +419,39 @@ v_d^i = v_d^r - R_d I_d
...
+PQ capability constraint at each terminal of a two-terminal VSC HVDC, added only
+on AC networks. For `HVDCTwoTerminalVSC` this is the exact circle:
+
+```math
+p_k^2 + q_k^2 \\le (S_k^{\\max})^2 \\quad k \\in \\{f, t\\}
+```
+
+For `HVDCTwoTerminalVSCMIP` it is replaced by an inscribed polygon to stay MILP.
+"""
+struct HVDCVSCReactiveCapabilityConstraint <: ConstraintType end
```
rodrigomha: This name is off. If this is a constraint for p^2+q^2, then this is constraint on ApparentPower, so it should be HVDCVSCApparentPowerLimitConstraint.

However a more important comment is that I don't understand why Claude is talking here about MIP or LP. This a circle constraint on apparent power, but we could skip this and only model the box constraints separately Pmin <= P <= Pmax and Qmin <= Q <= Qmax.

src/core/variables.jl
```diff
@@ -384,6 +387,34 @@ Docs abbreviation: ``z``
...
+abstract type To end
+
+"""
+Reactive power injected at the from- (`HVDCReactivePowerVariable{From}`) or to-
+(`HVDCReactivePowerVariable{To}`) terminal AC bus by a two-terminal HVDC link.
+Added only when the formulation runs on an AC network model.
+Docs abbreviation: ``q_f`` / ``q_t``
+"""
+struct HVDCReactivePowerVariable{Side} <: VariableType end
```
rodrigomha: Do you need to use the abstract From and To? Just define two structs for From and To. Using those abstracts is too much.

src/mt_hvdc_models/hvdcsystems_constructor.jl
```diff
@@ -54,16 +54,6 @@ function construct_device!(
     devices = get_available_components(model, sys)
     add_variables!(container, ActivePowerVariable, devices, T)
     add_variables!(container, ConverterCurrent, devices, T)
-    if _use_linear_loss(T, model)
```
rodrigomha, line 57: When these variables are being created?

src/mt_hvdc_models/hvdcsystems_constructor.jl
```diff
@@ -148,20 +140,23 @@ function construct_device!(
...
+    # The abs-value decomposition adds PositiveCurrent / NegativeCurrent
+    # variables that the ConverterLossConstraint reads, so it must run first.
+    #
+    # Note: use_linear_loss = true on QuadraticLossConverter adds a binary
+    # direction variable (CurrentDirection), making the model MINLP rather
+    # than smooth NLP. Caller is responsible for supplying a MINLP-capable
+    # solver in that case.
+    use_ll = get_attribute(model, "use_linear_loss")
+    if use_ll
+        _add_abs_value_decomposition!(
```
rodrigomha: Is this creating the variables in the ModelConstructor? I think we should try to create the variables in the ArgumentConstructor. Is fine if we have two auxiliary methods for constraints and variables separately.

src/network_models/pm_translator.jl
```diff
@@ -663,6 +663,27 @@ function get_branch_to_pm(
...
+function get_branch_to_pm(
+    ix::Int,
+    branch::PSY.TwoTerminalVSCLine,
+    ::Type{<:AbstractTwoTerminalVSCFormulation},
+    ::Type{<:AbstractPowerModel},
+)
+    return Dict{String, Any}()
+end
+
+# Trait: should this DeviceModel be skipped when translating two-terminal HVDC
```
rodrigomha, line 675: @jd-lara Is this right? Should we skip VSC Two Terminal to PowerModels since most of that is handled or our side right? We only use the generic for PM?

test/test_device_hvdc.jl
rodrigomha: This is good. Can you give this a shot to the water bilinear method that is more simple. We are using Bin2 there and it would be good to start trying to set-up an attribute that uses by default the bin2 one so we can start exploring like the dndt one

rodrigomha: You should branch out from the water bilinear to a new branch so we can merge in the meantime the other one.

src/common_models/quadratic_converter_loss.jl
```diff
@@ -0,0 +1,127 @@
...
+        JuMP.add_to_expression!(expr, b, i_neg_t)
+    end
+    return expr
+end
+
+#########################################
+####### Abs-value decomposition #########
+#########################################
+
+function _add_abs_value_decomposition!(
```
rodrigomha: I think we should split this method in two, for the variables and the constraints to keep consistency with the argument and model stages. Although that is more complex and more code since you will need to retrieve them in the other method, is the right way of doing it for the approach we use

src/ac_transmission_models/branch_constructor.jl
```diff
@@ -1670,3 +1670,180 @@ function construct_device!(
...
+    end
+
+    add_to_expression!(
+        container, ActivePowerBalance, FlowActivePowerFromToVariable,
+        devices, device_model, network_model,
+    )
+    add_to_expression!(
+        container, ActivePowerBalance, FlowActivePowerToFromVariable,
+        devices, device_model, network_model,
+    )
```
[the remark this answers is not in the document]
rodrigomha: @acostarelli Copilot is trolling here. the _maybe_add method also adds them to the expression. However, I would make the method in common files so it can be used with generic formulations like ThermalStandard or others. We do not need to refactor the old code to this, but future methods should try to use this.

src/twoterminal_hvdc_models/TwoTerminalDC_branches.jl
```diff
@@ -1433,3 +1433,513 @@ function add_constraints!(
...
+get_variable_upper_bound(::Type{NegativeCurrent}, d::PSY.TwoTerminalVSCLine, ::Type{<:AbstractTwoTerminalVSCFormulation}) = _vsc_shared_i_max(d)
+
+#! format: on
+
+####################### VSC reactive-power optional path #####################
+# The reactive power and PQ capability machinery is added only when the
+# network actually models reactive power. On active-only networks these
+# helpers are no-ops.
+
+function _maybe_add_reactive_power_variables!(
```
rodrigomha: @jd-lara What do you think of adding these methods a common approaches? This would help to consolidate the methods in the constructors to formulations that are basically the same except for the reactive power variable in AbstractPowerModel and AbstractActivePowerModel

jd-lara: I think it is ok. It helps reduce duplications

src/network_models/pm_translator.jl
jd-lara: for now yes until we move the PM models fully, I think it is better to keep current functionality and testing so that the transition is smoother.

rodrigomha (changes requested):
@acostarelli I think these comments are good starting point. Let me know when you address them and I will re-review.

Overall, is in a good shape. I have not reviewed the tests yet.

I think we should not use parametric types for now for variables or others, so let's revert that to multiple From/To structs specifically.

src/core/constraints.jl
acostarelli: I instructed it to make three formulations:
- NLP that uses the exact p^2+q^2
- MIP that approximates p^2 and q^2
- LP that builds an octagon (instead of a box)

src/common_models/quadratic_converter_loss.jl
acostarelli: Just so you know, the IOM approximation code sorta violates this because it creates all of its variables and constraints together, and we don't invoke the approximation code until the model stage. At least there it's all self-contained so POM doesn't technically "know" that IOM is making variables in the model stage.

src/twoterminal_hvdc_models/TwoTerminalDC_branches.jl
```diff
@@ -1433,3 +1433,472 @@ function add_constraints!(
...
+
+# DC voltage bounds (per-terminal)
+get_variable_lower_bound(::Type{HVDCFromDCVoltage}, d::PSY.TwoTerminalVSCLine, ::Type{<:AbstractTwoTerminalVSCFormulation}) = PSY.get_voltage_limits_from(d).min
+get_variable_upper_bound(::Type{HVDCFromDCVoltage}, d::PSY.TwoTerminalVSCLine, ::Type{<:AbstractTwoTerminalVSCFormulation}) = PSY.get_voltage_limits_from(d).max
+get_variable_lower_bound(::Type{HVDCToDCVoltage}, d::PSY.TwoTerminalVSCLine, ::Type{<:AbstractTwoTerminalVSCFormulation}) = PSY.get_voltage_limits_to(d).min
+get_variable_upper_bound(::Type{HVDCToDCVoltage}, d::PSY.TwoTerminalVSCLine, ::Type{<:AbstractTwoTerminalVSCFormulation}) = PSY.get_voltage_limits_to(d).max
+
+# Per-terminal accessors. Two concrete methods rather than one side-parametric
+# method to keep the public type surface narrow (no From/To tag types).
+_vsc_v_limits_from(d::PSY.TwoTerminalVSCLine) = PSY.get_voltage_limits_from(d)
```
acostarelli: I don't like these helpers. We already have the PSY routines for a reason.

src/ac_transmission_models/branch_constructor.jl
```diff
@@ -1670,3 +1670,176 @@ function construct_device!(
...
+# True when the PQ-capability constraint is `p_sq + q_sq <= s^2` using IOM's
+# quadratic-approx machinery (NLP exact + MIP PWL). False for the LP path,
+# which uses an octagonal linear outer-approximation instead and therefore
+# does not need p_sq / q_sq expressions.
+_uses_pq_sq_approx(::Type{HVDCTwoTerminalVSC}) = true
+_uses_pq_sq_approx(::Type{HVDCTwoTerminalVSCMILP}) = true
+_uses_pq_sq_approx(::Type{HVDCTwoTerminalVSCLP}) = false
+
+_vsc_v_bounds_from(devices) =
+    [(min = _vsc_v_limits_from(d).min, max = _vsc_v_limits_from(d).max) for d in devices]
```
acostarelli: The limits helper already returns a MinMax, why build another one?

src/common_models/add_to_expression.jl
```diff
@@ -718,6 +718,76 @@ function add_to_expression!(
     return
 end
 
+"""
+Two-terminal VSC: add `HVDCReactivePowerFromVariable` to `ReactivePowerBalance`
+at the from-terminal AC bus.
+"""
+function add_to_expression!(
```
acostarelli: We can still combine these two add_to_expression!s and dispatch on the variable type.

src/common_models/network_conditional.jl
```diff
@@ -0,0 +1,52 @@
...
+methods are responsible for the actual bus mapping and sign convention.
+"""
+function _maybe_add_reactive_power_variables!(
+    container::OptimizationContainer,
+    devices,
+    model::DeviceModel{D, F},
+    network_model::NetworkModel{N},
+    var_types,
+) where {D <: PSY.Device, F, N <: AbstractPowerModel}
+    if N <: AbstractActivePowerModel
```
acostarelli: This should be resolved using dispatch, not a conditional.

src/common_models/network_conditional.jl
```diff
@@ -0,0 +1,52 @@
...
+(`NetworkModel{<:AbstractPowerModel}`). No-op on active-power-only networks.
+"""
+function _maybe_add_reactive_power_constraints!(
+    container::OptimizationContainer,
+    devices,
+    model::DeviceModel{D, F},
+    network_model::NetworkModel{N},
+    constraint_type::Type{<:ConstraintType},
+) where {D <: PSY.Device, F, N <: AbstractPowerModel}
+    if N <: AbstractActivePowerModel
```
acostarelli: Same here

src/common_models/quadratic_converter_loss.jl
```diff
@@ -0,0 +1,141 @@
...
+        JuMP.add_to_expression!(expr, a, i_sq_t)
+    end
+    if use_linear_loss && !iszero(b)
+        JuMP.add_to_expression!(expr, b, i_pos_t)
+        JuMP.add_to_expression!(expr, b, i_neg_t)
+    end
+    return expr
+end
+
+function _quadratic_converter_loss_expr(
```
acostarelli, line 35: I feel like these functions can be combined in a smarter way. They're basically the same thing.

src/common_models/quadratic_converter_loss.jl
```diff
@@ -0,0 +1,141 @@
...
+    return
+end
+
+function _add_abs_value_decomposition_constraints!(
+    container::OptimizationContainer,
+    devices,
+    ::DeviceModel{D, F},
+    ::NetworkModel{<:AbstractPowerModel},
+    parent_var_type::Type{<:VariableType},
+    i_max_getter::Function,
```
acostarelli: Hmm I'm not a fan of this design requiring passing in a getter.

src/core/formulations.jl
```diff
@@ -184,8 +184,39 @@ Branch type to represent non-linear LCC (line commutated converter) model on two
...
+"""
+Two-terminal VSC formulation that keeps the bilinear ``v \\cdot I`` and quadratic
+``I^2`` terms exact. Requires an NLP-capable solver (e.g. Ipopt).
+"""
+struct HVDCTwoTerminalVSC <: AbstractTwoTerminalVSCFormulation end
+
+"""
+Two-terminal VSC formulation that approximates the bilinear ``v \\cdot I`` and
+quadratic ``I^2`` terms with SOS2-based piecewise-linear envelopes (the same
+scheme used by `MIPQuadraticLossConverter`). Stays MILP.
```
acostarelli: If the converter names haven't been updated, we're doing MILP instead of MIP in-case we also do MINLP-something one day.

src/twoterminal_hvdc_models/TwoTerminalDC_branches.jl
```diff
@@ -1433,3 +1433,472 @@ function add_constraints!(
...
+
+#! format: on
+
+####################### VSC PQ-approx registration ###########################
+# The generic `_maybe_add_reactive_power_variables!` /
+# `_maybe_add_reactive_power_constraints!` helpers live in
+# `common_models/network_conditional.jl` and are called directly from the VSC
+# constructor. This section only defines the VSC-specific
+# `_maybe_add_pq_sq_approx!` since it registers four named expressions that
+# are specific to the VSC variables (`p_ft_sq`, `p_tf_sq`, `q_f_sq`, `q_t_sq`).
```
acostarelli: Delete this comment.

src/twoterminal_hvdc_models/TwoTerminalDC_branches.jl
```diff
@@ -1433,3 +1433,472 @@ function add_constraints!(
...
+function _maybe_add_pq_sq_approx!(
+    container::OptimizationContainer,
+    devices,
+    line_names,
+    time_steps,
+    quad_cfg,
+    ::Type{F},
+    ::NetworkModel{<:AbstractPowerModel},
+) where {F <: AbstractTwoTerminalVSCFormulation}
+    _uses_pq_sq_approx(F) || return
```
acostarelli: This doesn't make any sense. If certain types should no-op, then they shouldn't be accepted by this method in the first place. This helper uses multiple dispatch to solve what multiple dispatch already solves.

src/twoterminal_hvdc_models/TwoTerminalDC_branches.jl
```diff
@@ -1433,3 +1433,472 @@ function add_constraints!(
...
+#   p_ft ==  v_f * I + (a_f * I^2 + b_f * |I| + c_f)
+#   p_tf == -v_t * I + (a_t * I^2 + b_t * |I| + c_t)
+# Sign convention: FlowActivePowerFromToVariable / ToFromVariable are positive
+# when the corresponding AC bus is sourcing power into the converter (matches
+# the existing add_to_expression! method's -1.0 multiplier).
+function add_constraints!(
+    container::OptimizationContainer,
+    ::Type{HVDCVSCConverterPowerConstraint},
+    devices::Union{
+        Vector{PSY.TwoTerminalVSCLine},
```
acostarelli: Is this the pattern that the rest of POM uses? To allow both a vector and the wrapper? And the visual pattern: to put the device in here and not the where clause?

src/twoterminal_hvdc_models/TwoTerminalDC_branches.jl
```diff
@@ -1433,3 +1433,472 @@ function add_constraints!(
...
+function add_constraints!(
+    container::OptimizationContainer,
+    ::Type{HVDCVSCApparentPowerLimitConstraint},
+    devices::Union{
+        Vector{PSY.TwoTerminalVSCLine},
+        IS.FlattenIteratorWrapper{PSY.TwoTerminalVSCLine},
+    },
+    ::DeviceModel{PSY.TwoTerminalVSCLine, F},
+    ::NetworkModel{<:AbstractPowerModel},
+) where {F <: Union{HVDCTwoTerminalVSC, HVDCTwoTerminalVSCMILP}}
```
acostarelli: Let's drop the MILP formulation and change the LP formulation as such: it builds first a box domain and then only makes it an octagon if an attribute is set to true, which it defaults to true. Make sure to add comments explaining this attribute. You can probably get rid of a bunch of stuff this way, like the maybe_pq stuff I don't know what that's about.

src/twoterminal_hvdc_models/TwoTerminalDC_branches.jl
```diff
@@ -1433,3 +1433,472 @@ function add_constraints!(
...
+        )
+    end
+
+    # Run the same octagon block twice (from/to) using a per-side spec, so
+    # the eight inner constraints aren't textually duplicated.
+    side_specs = (
+        (prefix = "from", p_var = p_ft, q_var = q_f, rating_getter = _vsc_rating_from),
+        (prefix = "to", p_var = p_tf, q_var = q_t, rating_getter = _vsc_rating_to),
+    )
+    for d in devices
```
acostarelli, line 1759: Can we reorganize this loop and use JuMP vectorized jump.@constraint macros to build this faster?

src/twoterminal_hvdc_models/TwoTerminalDC_branches.jl
```diff
@@ -1433,3 +1433,472 @@ function add_constraints!(
...
+# Their intersection is a regular octagon. Eight linear constraints per
+# terminal per timestep (four diagonal + four axis-aligned).
+#
+# Tightness: the octagon is loose by at most ≈8.2% in area (octagon-to-disk
+# area ratio = 8·tan(π/8)/π ≈ 1.082). It is *guaranteed* to be an outer
+# approximation: any solution feasible to the real disk is feasible to the
+# octagon, so the LP relaxation never blocks a physically realizable
+# operating point.
+#
+# Tradeoff vs the unified PWL path used by `HVDCTwoTerminalVSC` /
```
acostarelli: Simplify this now that we're dropping MILP.

src/twoterminal_hvdc_models/TwoTerminalDC_branches.jl
```diff
@@ -1433,3 +1433,472 @@ function add_constraints!(
...
+#
+#   • Axis-aligned square:  |p| ≤ rating, |q| ≤ rating
+#                           (tangent at (±rating, 0) and (0, ±rating))
+#   • 45°-rotated square:   |p| + |q| ≤ rating·√2
+#                           (tangent at the four diagonal points)
+#
+# Their intersection is a regular octagon. Eight linear constraints per
+# terminal per timestep (four diagonal + four axis-aligned).
+#
+# Tightness: the octagon is loose by at most ≈8.2% in area (octagon-to-disk
```
acostarelli: Are you sure it's an outer approximation?

acostarelli (commented):
Review 3

src/twoterminal_hvdc_models/TwoTerminalDC_branches.jl
rodrigomha: Agreed. These helpers exist because of the `{Side}` parametric design. Call the PSY accessors directly.

src/ac_transmission_models/branch_constructor.jl
rodrigomha: Ditto

src/common_models/add_to_expression.jl
rodrigomha: I think is fine having two methods. However, since the methods are super close I think you could reuse a single method with Union for both From and To, and having a helper method that retrieves the from or to bus depending on the variable. Proposed approach:

```julia

function add_to_expression!(
    ::Type{T},
    ::Type{U},
    devices::IS.FlattenIteratorWrapper{V},
    ::DeviceModel{V, W},
    network_model::NetworkModel{X},
) where {
    T <: ReactivePowerBalance,
    U <: Union{HVDCReactivePowerFromVariable, HVDCReactivePowerToVariable}
    V <: PSY.TwoTerminalVSCLine,
    W <: AbstractTwoTerminalVSCFormulation,
    X <: ACPPowerModel,
}
    var = get_variable(container, HVDCReactivePowerToVariable, V)
    nodal_expr = get_expression(container, T, PSY.ACBus)
    network_reduction = get_network_reduction(network_model)
    time_steps = get_time_steps(container)
    for d in devices
        name = PSY.get_name(d)
        bus = _get_bus_for_expression(d, U) # TODO: implement this helper method
        bus_no = PNM.get_mapped_bus_number(network_reduction, bus)
        for t in time_steps
            add_proportional_to_jump_expression!(
                nodal_expr[bus_no, t],
                var[name, t],
                -1.0,
            )
        end
    end
    return
end
```

Also, I think we should just add the ReactivePowerVariable positive to the expression. This is mostly a convention for VSC endings, that they can inject/consume reactive power freely, so it's fine to have them positive (contrary to the LCC model that has to consume reactive power).

src/common_models/network_conditional.jl
rodrigomha: Yep, super weird Claude here lol

src/common_models/quadratic_converter_loss.jl
rodrigomha: These two methods are the correct approach, we should have the methods because of the returning expression being Quad or Aff. We should not combine these methods. However, similar to my previous code a helper method could potentially be used to define correctly the expression, but the methods are super short so adding the helper method is probably not a good idea and will end up being more confusing.

src/common_models/quadratic_converter_loss.jl
rodrigomha: Yeah, this is pretty bad, don't pass a function. Ask Claude to refactor to dispatch on device type. Helper functions are okay if the PSY accesors for `I_max` are different.

src/twoterminal_hvdc_models/TwoTerminalDC_branches.jl
rodrigomha: Right? Because of the method of L1567?

src/twoterminal_hvdc_models/TwoTerminalDC_branches.jl
rodrigomha: It seems that both approaches are somewhat used depending on which methods expect only the wrapper or both (wrapper or vector). So It's fine overall to have this.

src/twoterminal_hvdc_models/TwoTerminalDC_branches.jl
rodrigomha: You can ask Claude to try this, but this is okay as it is. Note that we are adding a container for each tag, so it can be a pain to implement that.

src/twoterminal_hvdc_models/TwoTerminalDC_branches.jl
rodrigomha: I asked Claude to make a nice figure of this, but yeah, pretty much, since the diagonal is rating * sqrt(2.0).

image

src/ac_transmission_models/branch_constructor.jl
```diff
@@ -1670,3 +1670,158 @@ function construct_device!(
...
+    add_variables!(container, HVDCFromDCVoltage, devices, F)
+    add_variables!(container, HVDCToDCVoltage, devices, F)
+
+    _maybe_add_reactive_power_variables!(
+        container, devices, device_model, network_model,
+        (HVDCReactivePowerFromVariable, HVDCReactivePowerToVariable),
+    )
+
+    # use_linear_loss = true adds a binary direction variable (CurrentDirection),
+    # making the model MINLP rather than smooth NLP. Caller is responsible for
```
acostarelli: Is this exclusively for NLPs? Does the MILP branch not go through here as well?

src/common_models/quadratic_converter_loss.jl
acostarelli: What do you think of this setup with the helper? @rodrigomha

src/mt_hvdc_models/hvdcsystems_constructor.jl
```diff
@@ -54,15 +54,14 @@ function construct_device!(
...
-            @warn "use_linear_loss is enabled but every InterconnectingConverter has a zero proportional loss term; no linear-loss variables/constraints will be added."
-        else
-            add_variables!(container, ConverterPositiveCurrent, ll_devices, T)
-            add_variables!(container, ConverterNegativeCurrent, ll_devices, T)
-            add_variables!(container, ConverterCurrentDirection, ll_devices, T)
-        end
+
+    # use_linear_loss = true on QuadraticLossConverter adds a binary direction
+    # variable (CurrentDirection), making the model MINLP rather than smooth
+    # NLP. Caller is responsible for supplying a MINLP-capable solver.
```
acostarelli: Again, doesn't MILP also dispatch through here?

src/twoterminal_hvdc_models/TwoTerminalDC_branches.jl
```diff
@@ -1433,3 +1433,402 @@ function add_constraints!(
...
+# Register the four IOM `QuadraticExpression` handles (`p_ft_sq`, `p_tf_sq`,
+# `q_f_sq`, `q_t_sq`) that the disk constraint reads. Only the NLP formulation
+# on an AC network actually emits the disk constraint, so only that combo
+# registers anything; the LP path and active-power-only networks no-op.
+function _register_pq_sq_expressions!(
+    container::OptimizationContainer,
+    devices,
+    line_names,
+    time_steps,
+    quad_cfg,
```
acostarelli: Why is the quad config being passed in through argument? We know this is the NLP dispatch.

src/twoterminal_hvdc_models/TwoTerminalDC_branches.jl
```diff
@@ -1433,3 +1433,402 @@ function add_constraints!(
...
+    ::DeviceModel{PSY.TwoTerminalVSCLine, HVDCTwoTerminalVSC},
```
acostarelli: I think it's maybe confusing that the concrete type looks a lot like the abstract type here. Maybe we make the concrete type also say NLP.

src/twoterminal_hvdc_models/TwoTerminalDC_branches.jl
```diff
@@ -1433,3 +1433,402 @@ function add_constraints!(
...
+    )
+    for d in devices
+        name = PSY.get_name(d)
+        for spec in side_specs
+            rating = spec.rating_getter(d)
+            diag = rating * sqrt(2.0)
+            prefix = spec.prefix
+            p_var, q_var = spec.p_var, spec.q_var
+            for t in time_steps
+                cons[prefix * "_p_ub"][name, t] =
```
acostarelli, line 1767: If we didn't this prefix system, could we simplify this using the vectorized constraints?

acostarelli (commented):
Review 4

src/twoterminal_hvdc_models/TwoTerminalDC_branches.jl
```diff
@@ -1433,3 +1433,362 @@ function add_constraints!(
...
+        # If g == 0 we treat the cable as lossless (v_f == v_t).
+        # Otherwise R = 1/g.
+        for t in time_steps
+            cons[name, t] = if iszero(g)
+                JuMP.@constraint(jump_model, v_f[name, t] == v_t[name, t])
+            else
+                JuMP.@constraint(
+                    jump_model,
+                    v_f[name, t] - v_t[name, t] == (1.0 / g) * i_var[name, t],
+                )
```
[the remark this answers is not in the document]
acostarelli: @rodrigomha Does this look right now?

src/core/formulations.jl
```diff
@@ -197,7 +197,7 @@ abstract type AbstractTwoTerminalVSCFormulation <: AbstractTwoTerminalDCLineForm
 Two-terminal VSC formulation that keeps the bilinear ``v \\cdot I`` and quadratic
 ``I^2`` terms exact. Requires an NLP-capable solver (e.g. Ipopt).
 """
-struct HVDCTwoTerminalVSC <: AbstractTwoTerminalVSCFormulation end
+struct HVDCTwoTerminalVSCNLP <: AbstractTwoTerminalVSCFormulation end
```
acostarelli, line 200: @rodrigomha What do you think of this naming?

src/common_models/quadratic_converter_loss.jl
rodrigomha: Sure... is fine, is not that terrible indirection and looks cleaner

src/core/formulations.jl
rodrigomha: It's okay I think. I'm thinking that maybe NLPHVDCTwoTerminalVSC and LPHVDCTwoTerminalVSC is better, but I don't know lol. This approach is more consistent with the Converter convention using MILPQuadraticLossConverter.

Any ideas @jd-lara

src/twoterminal_hvdc_models/TwoTerminalDC_branches.jl
rodrigomha: This is good for future testing but not really needed for this PR, so we can keep it as it is for now, but an issue could be open if you want to explore the idea

src/ac_transmission_models/branch_constructor.jl
```diff
@@ -1670,3 +1670,161 @@ function construct_device!(
...
+
+############################################################################
+####################### Two-Terminal VSC HVDC Construct ####################
+############################################################################
+
+# Quadratic / bilinear approximation traits — same scheme used by the MT
+# converter formulations.
+_quad_config(::Type{HVDCTwoTerminalVSCNLP}) = IOM.NoQuadApproxConfig()
+_quad_config(::Type{HVDCTwoTerminalVSCLP}) =
+    IOM.SolverSOS2QuadConfig(DEFAULT_INTERPOLATION_LENGTH)
```
rodrigomha, line 1682: @acostarelli Can you open an issue to figure out how the user can configure the interpolation and model option (BIN2 or other) for the approximation for the future? We don't need to solve this in this PR, but we need to be aware of that.

src/twoterminal_hvdc_models/TwoTerminalDC_branches.jl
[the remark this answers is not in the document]
rodrigomha: This is correct. If g = 0, then there is an infinite impedance and hence cannot be current flowing through the cable. Good catch here.

rodrigomha (approved):
Okay. I'm happy with this. We can merge this @jd-lara after we decide if we are okay with the naming convention

PR #124, opened 2026-05-15
Remove model from being interpolated into warning strings

odow:
First one is a bug.

The others: interpolating a `built!` model is very expensive. I was running `POM.solve!(model)` and wondering why there was a 5 minute delay and so much stuff being printed to the screen. I don't know if removing is okay, or if there is a better identifier we can use here.


jd-lara (changes requested):
We can even convert that warn into a @debug call

jd-lara (approved):

jd-lara (approved):

PR #135, opened 2026-06-09
Consolidate add_to_expression! device-injection methods; fix AreaBalance bugs

luke-kiernan:
## Summary
In the initial PSI into IOM+POM split, the 90+ variations on `add_to_expression!` was one thing where I told myself "I'll fix this properly later." This PR is to get the ball rolling on that fix: refactor so as to minimize code spaghetti and ensure good test coverage.

Consolidates the device-injection family of `add_to_expression!` methods onto a shared, network-model-aware target resolver (`_balance_expression_targets`) plus IOM's new `add_device_terms_to_expression!` driver. ~17 hand-rolled `device × time` loops collapse to thin delegations. Net **−81 lines** in `add_to_expression.jl`.

Depends on IOM PR Sienna-Platform/InfrastructureOptimizationModels.jl#112 (the driver), which targets IOM `main`.

## What's consolidated
- **variable → balance** (StaticInjection), across nodal / AreaBalance / CopperPlate / PTDF / AreaPTDF.
- **parameter → balance** (TimeSeriesParameter, ElectricLoad TS-param, MotorLoad constant, OnStatusParameter) for nodal / AreaBalance / CopperPlate.
- **compact-UC `OnVariable`**: a single resolver-based `_add_compact_on_to_balance!` (must-run aware) replaces the nodal-only `_add_to_expression!` and both AreaBalance variants; CopperPlate/PTDF use a new `scale` kwarg on `_add_variable_to_balance!` for the per-device `p_min`.

PTDF *parameter* methods are intentionally left out (their `AbstractPTDFModel`/security-constrained semantics differ from the variable family and need a separate decision).

## Bug fixes (port of PSI `22d2059c2`, which POM missed by branching before it — see #134)
- **must-run compact-UC `OnVariable`** called a non-existent 2-arg `add_proportional_to_jump_expression!` → `MethodError`. Now `add_constant_to_jump_expression!` (On≡1 → P_min is a constant).
- **`OnVariable`/`ThermalGen` + AreaBalance** indexed an `ACBus` container AreaBalance never creates → `KeyError`. Now `PSY.Area`.
- **`FlowActivePowerToFromVariable`/`TwoTerminalHVDC` + AreaBalance** used an undefined `network_reduction` and unbound `W` → `UndefVarError`. Now Area-indexed, `HVDCTwoTerminalDispatch`.
- **compact-UC `OnVariable` under AreaBalance/AreaPTDF**: routing through the resolver fixes a `KeyError` (AreaBalance) and a silent missing-area-contribution (AreaPTDF, where the generic method previously added to nodal only).

## Type stability / perf
The driver's inner loop is allocation-free even on heterogeneous target tuples (AreaPTDF mixes a `String`-keyed area expression and an `Int`-keyed nodal expression) via `Base.tail` recursion. Cold build time unchanged.

## Tests
Constructor suites pass against IOM `main` + the driver branch (thermal/load/renewable/network/branch/HVDC; 5244/5244 on the thermal+network subset). The fixed paths above are currently unexercised by any test system — tracked in #134 for follow-up coverage.

## Follow-ups
- This PR pins IOM to `lk/add-device-terms-driver` (= IOM `main` + the driver) so it builds now. After IOM #112 merges into `main`, flip the two `[sources]` pins here back to `rev = "main"`.
- Add the two AreaBalance/AreaPTDF compact spots to #134.


src/common_models/add_to_expression.jl
```diff
@@ -14,6 +14,255 @@ _system_expression_type(::Type{AreaPTDFPowerModel}) = PSY.Area
...
+unit-commitment `OnVariable`).
+"""
+function _add_variable_to_balance!(
+    container::OptimizationContainer,
+    ::Type{T},
+    ::Type{U},
+    devices::Union{Vector{V}, IS.FlattenIteratorWrapper{V}},
+    network_model::NetworkModel,
+    ::DeviceModel{V, W};
+    scale::Union{Nothing, Function} = nothing,
```
luke-kiernan: iirc positional args are better than keyword args for compilation purposes

src/common_models/add_to_expression.jl
```diff
@@ -14,6 +14,255 @@ _system_expression_type(::Type{AreaPTDFPowerModel}) = PSY.Area
...
+# device/time loop. Dispatch is on disjoint network-model subtrees, so no ambiguity:
+#   - <:AbstractPowerModel        -> nodal bus only      (default AC models; also the
+#                                                         security-constrained PTDF
+#                                                         variants, matching prior
+#                                                         variable-method behavior)
+#   - CopperPlatePowerModel       -> system reference bus only
+#   - AreaBalancePowerModel       -> area only
+#   - PTDFPowerModel/AreaPTDF...  -> system/area + nodal bus
+#################################################################################
+
```
luke-kiernan, line 32: Hmm. Dramatically better, but still some redundacy here.
1. First thing in return value in `_balance_expression_targets` is always `get_expression(container, T, [something])`
2. Final version of `_balance_expression_targets` is basically combination of nodal + area. Also it's the only spot where `_system_expression_type` is called anywhere in this file.
3. All the `_add_foo_to_balance!` functions only differ in what they pass as the 3rd argument: 
```julia
add_device_terms_to_expression!(container,
        d -> _balance_expression_targets(container, T, network_model, d),
        function (d)
           ... # this differs
        end,
        devices,
 )
```
That being said, I'm not sure how feasible it is to captialize on this: we do want the `get_{variable/multiplier/parameter}` calls to happen once, not per-device.

src/common_models/add_to_expression.jl
```diff
@@ -14,6 +14,255 @@ _system_expression_type(::Type{AreaPTDFPowerModel}) = PSY.Area
...
+        d -> _balance_expression_targets(container, T, network_model, d),
+        function (d)
+            name = PSY.get_name(d)
+            multiplier = PSY.get_active_power_limits(d, PSY.SU).min * base_multiplier
+            if PSY.get_must_run(d)
+                # On ≡ 1 for must-run units, so the term is the constant p_min * mult.
+                return t -> (1.0, multiplier)
+            else
+                return t -> (variable[name, t], multiplier)
+            end
```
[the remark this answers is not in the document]
luke-kiernan: Issue opened for this already: #134. I want feedback on the whole concept of this PR then will add tests.

PR #138, opened 2026-06-15
Add time-varying ORDC; port SC slacks

acostarelli:
Port of PowerSimulations.jl#1629 (time-varying ORDC), adapted to the psy6 data model where a time-varying ORDC is a `ReserveDemandTimeSeriesCurve` whose `variable` is a `CostCurve{TimeSeriesPiecewiseIncrementalCurve}` (rather than a `ReserveDemandCurve` carrying a bare `TimeSeriesKey`).

- add_parameters.jl: service-level piecewise slope/breakpoint cost-parameter population for `ReserveDemandTimeSeriesCurve` (calc_additional_axes, get_max_tranches, _get_time_series_name, _add_parameters!, _param_to_vars).
- reserves.jl / market_bid_overrides.jl: drop the "not supported" error and route ORDC cost through the decremental block-offer path, reading per-timestep data from the parameter arrays (time-varying) or the curve (static); time-varying cost goes to the variant objective expression.
- services_constructor.jl: route time-series-backed ORDC to parameter setup.
- Broaden shared reserve / production-cost-expression methods to Union{ReserveDemandCurve, ReserveDemandTimeSeriesCurve}; update hybrid/storage disambiguators to keep dispatch unambiguous.
- test/test_services_constructor.jl: build, and build & solve, time-varying ORDC.
- Fix a stray `end` in hybrid_systems.jl that blocked precompilation.
- Point the IOM source at the ac/ordc branch (companion changes live there).

Thanks for opening a PR to PowerOperationsModels.jl, please take note of the following when making a PR:

Check the [contributor guidelines](https://sienna-platform.github.io/PowerOperationsModels.jl/stable/api/developer_guidelines/)


luke-kiernan (changes requested):
On the "port former PNM methods" side, can someone explain to me why these things were removed from PNM in the first place? imo they'd belong better back there.

On the ORDC side: _**very little of the code I'm seeing is ORDC specific**_. 95% of it applies equally well to other PWL offer curves. That tells me we should be calling helpers back in IOM instead of copy-pasting. I had Claude code take a look:

> 1. `_add_parameters!` should be unified. The device `ObjectiveFunctionParameter` method already handles N components; pass the service as a 1-element collection.
> 2. `_get_reserve_pwl_data` should route through `OfferDirection`, like the other 3 consumers of the PWL offer abstraction (gen incremental, load decremental, and market-bid). Right now, it re-derives the TS/static split and hard codes `Decremental*`. Same for the block-offer var/constraint selection in `add_pwl_term_delta!`.
> 3. The hardcoded `Decremental*` types mean the reserve path silently can't ever be incremental; going through `_slope_param(dir)` makes that a one-line dispatch rather than a structural assumption.

src/common_models/add_parameters.jl
```diff
@@ -495,31 +532,80 @@ calc_additional_axes(
...
-    U <: Union{Vector{D}, IS.FlattenIteratorWrapper{D}},
-    W <: AbstractDeviceFormulation,
-} where {D <: PSY.Component}
+    ::Type{T},
+    devices::V,
+    model,
+    ::Type{W};
+    meta = IOM.CONTAINER_KEY_EMPTY_META,
+) where {T <: ObjectiveFunctionParameter, V, W}
+    D = eltype(devices)
```
acostarelli, line 608: Why do eltype when we can specify D and W in the device model

src/common_models/add_parameters.jl
```diff
@@ -495,31 +532,80 @@ calc_additional_axes(
 } where {D <: PSY.Service} = ()
 
 #################################################################################
-# _unwrap_for_param — default implementation (identity)
+# Piecewise linear parameter helpers (time-varying ORDC)
 #################################################################################
 
-_unwrap_for_param(::ParameterType, ts_elem, expected_axs) = ts_elem
+function get_max_tranches(component::PSY.Component, piecewise_ts::IS.TimeSeriesKey)
```
acostarelli, line 538: Maybe the naming is weird here now with IOM having the same but this not overriding it.

src/common_models/add_parameters.jl
```diff
@@ -379,6 +395,19 @@ _get_time_series_name(
...
+_get_time_series_name(
+    ::Type{
+        <:Union{
+            DecrementalPiecewiseLinearSlopeParameter,
+            DecrementalPiecewiseLinearBreakpointParameter,
+        },
+    },
+    service::PSY.ReserveDemandTimeSeriesCurve,
+    ::ServiceModel,
+) = IS.get_name(IS.get_time_series_key(PSY.get_variable(service)))
```
acostarelli, line 409: This body doesn't feel specific to the parameters or service at all, can we loosen the types on another method?

src/common_models/add_parameters.jl
```diff
@@ -467,6 +496,14 @@ _param_to_vars(
     ::Type{<:AbstractDeviceFormulation},
 ) =
     (PiecewiseLinearBlockDecrementalOffer,)
+_param_to_vars(
```
acostarelli, line 499: Maybe we could give all of these helpers a more descriptive name while we're here.

src/common_models/add_parameters.jl
acostarelli: Also we should type all of these parameters

src/network_models/instantiate_network_model.jl
```diff
@@ -173,9 +173,125 @@ function _get_irreducible_buses_due_to_monitored_components(
...
+# arc in `name_to_arc_maps`, so building its flow vars/constraints would fail. Absence
+# from the map alone is not enough: types that never use it (e.g. HVDC) are also
+# absent, so we prune only when an endpoint bus was actually removed by the reduction.
+# Uncommon; `model_all_branches` keeps such lines instead.
+function _prune_fully_reduced_branch_models!(
+    network_model::NetworkModel,
+    branch_models::BranchModelContainer,
+)
+    merged_buses = Set{Int64}()
+    for removed in values(PNM.get_bus_reduction_map(network_model.network_reduction))
```
luke-kiernan, line 219: First, I believe the buses removed in a degree 2 reduction are stored separately, in `removed_buses`. This should probably consider those too.

Second, iterating over `values` of the bus reduction map means the criteria is directional: "it's a problem when the bus is absorbed in the merge; it's ok if the bus aborbs others." That's a red flag to me. Concretely: if we have a monitored line, I believe the current thru the arc of the monitored line is altered by the network reduction _in both situations_. So both "was absorbed" and "absorbed others" should be flagged.

Maybe we should probably just flag the buses adjacent to monitored lines as ineligible for network reductions (at least degree 2 and radial: maybe zero-impedance is ok) back in PNM.

luke-kiernan: Okay this seems complex enough that I opened an issue for it back in PNM: [#328](https://github.com/Sienna-Platform/PowerNetworkMatrices.jl/issues/328)

luke-kiernan (approved):
Looks like a fair port of the PSI PRs. Reading through this carefully raised some questions/concerns about the interaction between monitored lines and network reductions, but (1) it seems like it's mostly fine after all and (2) it can be discussed/addressed separately.

edit: okay I am a little troubled by the fact that the tests aren't passing. Just env issues?

jd-lara (approved):

PR #141, opened 2026-06-23
Complete the device time-varying market-bid offer-curve path

rodrigomha:
## Summary

Completes the **device** time-varying market-bid offer-curve path. The time-varying PWL market-bid machinery was wired for **ORDC services** only; a `ThermalStandard` (or other device) carrying a `MarketBidTimeSeriesCost` cascaded through several missing device-side methods during `build!`. This adds the device counterparts so a 24h time-varying offer-curve UC builds and solves.

Pairs with Sienna-Platform/InfrastructureOptimizationModels.jl#`rh/scalar-ts-cost-unwrap` (scalar `unwrap_for_param`), which fixes the first crash in the chain.

## What was missing (each surfaced as a separate `build!` failure)

1. **`_get_time_series_name`** for `Incremental/DecrementalPiecewiseLinearSlope/BreakpointParameter` on a device — the generic `get_time_series_names(model)[T]` has no entry → `KeyError`. Added device methods reading the offer-curve time-series key from the device's `MarketBidTimeSeriesCost` (analogous to the existing ORDC-service method).
2. **`calc_additional_axes`** for those PWL params on a device — the default device method returns `()`, so the parameter container was built with no tranche axis and the PWL element fell through `unwrap_for_param` → `size(::PiecewiseStepData)`. Added device methods that size the container to the batch-wide max tranches (via `get_max_tranches`), padding shorter per-hour curves.
3. **`_get_cost_if_exists(::MarketBidTimeSeriesCost) = nothing`** — the generic falls to `get_variable(cost)` (a service/fuel accessor) → `MethodError`. Mirrors the existing static `MarketBidCost` method (TS bids aren't fuel curves).

All additions are small and additive, mirroring the existing static-`MarketBidCost` / ORDC-service handling; no changes to existing methods.

## Verification

A 24h copper-plate UC with `ThermalStandard` gens (and an `InterruptiblePowerLoad` bid) carrying time-varying offer curves now `build!`s and `solve!`s, with the per-hour JuMP block-offer objective coefficients matching the per-hour offer-curve slopes exactly. Verified on both synthetic data and ~1 month of real ERCOT DAM offer curves (per-hour `SingleTimeSeries` → `transform_single_time_series!` → 24h `Deterministic` window).

Note: building a one-sided device cost (supply-only gen / demand-only load) requires marking the unused side with an empty-name time-series key (the placeholder the validation keys off). PSY exposes no helper for this — worth a follow-up to add one, analogous to the static `ZERO_OFFER_CURVE`.


src/common_models/add_parameters.jl
```diff
@@ -459,6 +459,49 @@ function _get_time_series_name(
...
+    op_cost = PSY.get_operation_cost(device)
+    IS.@assert_op op_cost isa TS_OFFER_CURVE_COST_TYPES
+    return IS.get_name(
+        IS.get_time_series_key(
+            PSY.get_value_curve(PSY.get_incremental_offer_curves(op_cost)),
+        ),
+    )
+end
+
+function _get_time_series_name(
```
acostarelli: We can combine these with a helper.

src/common_models/add_parameters.jl
```diff
@@ -570,6 +613,54 @@ function calc_additional_axes(
...
+# Device-side counterparts of the ORDC service methods above: the PWL slope &
+# breakpoint axes for a time-varying device offer curve come from the offer-curve
+# time series referenced by the device's `MarketBidTimeSeriesCost`. The container
+# is sized to the batch-wide maximum number of tranches (shorter per-hour curves
+# are padded in `unwrap_for_param`). Without these, the default device method
+# returns `()` and the PWL element falls through `unwrap_for_param` to `size(...)`.
+function _device_offer_curve_ts_key(::Type{T}, device::PSY.Component) where {T <: ParameterType}
+    op_cost = PSY.get_operation_cost(device)
+    IS.@assert_op op_cost isa TS_OFFER_CURVE_COST_TYPES
+    curve = if T <: Union{
```
acostarelli: We should dispatch on this instead.

src/common_models/add_parameters.jl
```diff
@@ -459,6 +459,49 @@ function _get_time_series_name(
     )
 end
 
+# Time-varying device offer curves: the incremental/decremental PWL slope &
```
acostarelli: I think this comment is excessive, the other methods don't use documentation.

src/common_models/add_parameters.jl
acostarelli: There are also other _get_time_series_name methods with similar logic so maybe this could use a larger refactor.

acostarelli (changes requested):

jd-lara (approved):

PR #147, opened 2026-06-25
[WIP] Add Sienna PSI→POM port plan to .claude

acostarelli:
I'm imagining using this branch to track remaining things to port and accumulating the remaining ports.

In addition, we should run a coverage check.


jd-lara (commented):
This is ok temporarily while we do the transition

PR #148, opened 2026-06-26
Port G-1 security-constrained reserves (PSI #1617)

acostarelli:
https://github.com/Sienna-Platform/PowerSimulations.jl/pull/1617/


src/core/expressions.jl
```diff
@@ -6,7 +6,11 @@ struct ComponentReserveUpBalanceExpression <: ExpressionType end
 struct ComponentReserveDownBalanceExpression <: ExpressionType end
 struct InterfaceTotalFlow <: ExpressionType end
 struct PTDFBranchFlow <: ExpressionType end
+abstract type PostContingencySystemBalanceExpressions <: SystemBalanceExpressions end
+struct PostContingencyActivePowerBalance <: PostContingencySystemBalanceExpressions end
+struct PostContingencyAreaInterchangeFlow <: PostContingencyExpressions end
```
acostarelli, line 11: I think PostContingencyExpressions is an IOM type. Maybe it should move, along with the other contingency types it has.

test/test_static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,2592 @@
+function get_outage_total_power_by_step_dict(
```
acostarelli, line 1: Double-check that all test cases were ported as 1:1 as possible from the original.

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1804 @@
...
+function construct_service!(
+    container::OptimizationContainer,
+    sys::PSY.System,
+    ::ModelConstructStage,
+    model::ServiceModel{SR, SecurityConstrainedRampReserve},
+    ::Dict{Symbol, DeviceModel},
+    ::Set{<:DataType},
+    network_model::NetworkModel{<:AreaBalancePowerModel},
+) where {SR <: PSY.AbstractReserve}
+    _construct_service_model_areabalance!(
```
acostarelli: For these `construct_service!` methods, we have these `_construct_service_model_*` helpers when we could instead dispatch on the network model.

acostarelli: It also seems like those helpers have a lot of similarities.

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1804 @@
...
+            ] for (uuid, entries) in resolved
+        ]
+    slack_ub = if use_slacks
+        add_post_contingency_slack_variables!(
+            container,
+            PostContingencyFlowActivePowerSlackUpperBound,
+            service,
+            service_name,
+            slack_resolved,
+            F(),
```
acostarelli: We should be passing by type not instance.

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1804 @@
...
+    resolved = _resolve_service_monitored_area_interchanges(sys, service_model)
+
+    contents = Dict{Tuple{String, String, Int}, JuMP.AffExpr}()
+    for (uuid, entries) in resolved
+        outage_id = string(uuid)
+        for (name, _) in entries, t in time_steps
+            contents[(outage_id, name, t)] = zero(JuMP.AffExpr)
+        end
+    end
+    expression_container = SparseAxisArray(contents)
```
acostarelli: I'm seeing this a lot in this PR, creating a SparseAxisArray directly and then the keys manually. Check the IOM API, we might be able to do this cleaner than we were doing in PSI. If not, we should probably update the IOM API.

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1804 @@
...
+    con_ub = _add_service_post_contingency_sparse_constraints!(
+        container, T, R, service_name; meta_suffix = "ub",
+    )
+    isempty(resolved) && return
+
+    post_cont_flow = get_expression(container, U, R, service_name)
+    jump_model = get_jump_model(container)
+
+    use_slacks = get_use_slacks(service_model)
+    slack_ub = if use_slacks
```
acostarelli: I've seen this use_slacks routine at least in one other place.

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1804 @@
...
+) = 1.0
+get_variable_multiplier(
+    ::Type{<:VariableType},
+    ::Type{<:PSY.Generator},
+    ::Type{<:AbstractSecurityConstrainedReservesFormulation},
+) = -1.0
+#! format: on
+
+# Add `multiplier * var` to a JuMP scalar in place (post-contingency builders).
+function _add_to_jump_expression!(
```
acostarelli: This already exists in jump_utils.jl

test/test_static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,2592 @@
...
+        devices = PSY.get_associated_components(
+            sys,
+            outage;
+            component_type = PSY.Generator,
+        )
+        for (i, device) in enumerate(devices)
+            device_name = PSY.get_name(device)
+            current_v =
+                filter(x -> x[col_name] == device_name, required_variables)[!, "value"]
+            if i == 1
```
acostarelli: Surely there's a more Julian way to do this sum.

acostarelli (commented):
Review 1

src/core/expressions.jl
acostarelli: Leave this for later.

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1635 @@
...
+
+function construct_service!(
+    container::OptimizationContainer,
+    sys::PSY.System,
+    ::ModelConstructStage,
+    model::ServiceModel{SR, F},
+    ::Dict{Symbol, DeviceModel},
+    ::Set{<:DataType},
+    network_model::NetworkModel{
+        <:Union{PM.AbstractDCPModel, CopperPlatePowerModel, AreaBalancePowerModel},
```
acostarelli: Do we have to rely on PowerModels here? We are trying to phase out this dependency, but maybe we still need it right now.

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1635 @@
...
+    add_to_expression!(
+        container, sys, PostContingencyAreaActivePowerDeployment, ActivePowerVariable,
+        service, model, network_model,
+    )
+    add_constraints!(
+        container, sys, PostContingencyCopperPlateBalanceConstraint,
+        PostContingencyAreaActivePowerDeployment, ActivePowerBalance,
+        service, model, network_model,
+    )
+    add_post_contingency_flow_expressions!(
```
acostarelli, line 1574: Reminder for myself: why do we have multiple dispatches, one that takes sys and one that doesn't? If we passed sys anyway could we combine with the PTDF one?

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1635 @@
...
+        PostContingencyActivePowerBalance,
+        contributing_devices, service, model, network_model,
+    )
+    return attribute_device_map
+end
+
+# Shared ModelConstructStage skeleton for every supported network model. The
+# network-specific post-contingency deployment/flow terms are dispatched on the
+# network model through `_add_post_contingency_network_terms!`.
+function _construct_service_model_sc!(
```
acostarelli: This does not need to be a separate function...

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1635 @@
...
+    SR <: PSY.AbstractReserve,
+    F <: AbstractSecurityConstrainedReservesFormulation,
+}
+    add_to_expression!(
+        container, sys, PostContingencyActivePowerBalance,
+        PostContingencyActivePowerReserveDeploymentVariable,
+        contributing_devices, service, model, network_model,
+    )
+    attribute_device_map = PSY.get_component_supplemental_attribute_pairs(
+        PSY.Generator, PSY.UnplannedOutage, sys,
```
acostarelli: What is the difference between UnplannedOutages and regular Outages? Should this dispatch on an abstract type?

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1635 @@
...
+
+# ----------------------------------------------------------------------------
+# construct_service! dispatches: argument + model construct stages for
+# (SecurityConstrainedContingencyReserve, SecurityConstrainedRampReserve) ×
+# (PTDF, CopperPlate, AreaBalance).
+# ----------------------------------------------------------------------------
+
+# Shared ArgumentConstructStage helper used by both formulations: builds
+# pre-contingency reserve variable + post-contingency deployment variable.
+function _construct_service_arguments_sc!(
```
acostarelli: This also doesn't need to be a separate function.

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1635 @@
...
+    end
+
+    add_variables!(
+        container,
+        sys,
+        PostContingencyActivePowerReserveDeploymentVariable,
+        service,
+        model,
+        contributing_devices,
+        F(),
```
acostarelli: We should dispatch on type not instance

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1635 @@
...
+                    jump_model, expressions[outage_id, name, t] >= limits.min,
+                )
+            end
+        end
+    end
+    return
+end
+
+"""
+Per-(outage, area, t) area balance for the `AreaBalancePowerModel`: the
```
acostarelli: Unnecessary comment

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1635 @@
...
+                    variable_outage[outage_id, name, t] <= variable[name, t],
+                )
+            end
+        end
+    end
+    return
+end
+
+"""
+Per-(outage, generator, t) min/max bounds on the
```
acostarelli: Mostly unnecessary comment, the last sentence is useful.

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1635 @@
...
+                _add_post_contingency_flow_rate_constraint!(
+                    jump_model, con_ub, con_lb, post_cont_flow, slack_ub, slack_lb,
+                    outage_id, name, t, lb, ub,
+                )
+            end
+        end
+    end
+    return
+end
+
```
acostarelli, line 1114: How do contingencies affect the ActivePowerBalance expression?

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1635 @@
...
+    resolved = _resolve_service_monitored_area_interchanges(sys, service_model)
+
+    index_keys = [
+        (string(uuid), name, t)
+        for (uuid, entries) in resolved for (name, _) in entries for t in time_steps
+    ]
+    expression_container = IOM.sparse_container_spec(JuMP.AffExpr, index_keys)
+    IOM._assign_container!(
+        container.expressions,
+        ExpressionKey(T, R, service_name),
```
acostarelli: Again we shouldn't be having to manually create keys. Either we're not using the API correctly, or the API needs to change. How are other sparse containers in POM constructed?

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1635 @@
...
+    for outage in associated_outages
+        associated_devices =
+            PSY.get_associated_components(sys, outage; component_type = PSY.Generator)
+        outage_id = string(IS.get_uuid(outage))
+        for device in contributing_devices
+            mult = device in associated_devices ? 0.0 : mult_default
+            name = PSY.get_name(device)
+            area_name = PSY.get_name(PSY.get_area(PSY.get_bus(device)))
+            for t in time_steps
+                JuMP.add_to_expression!(
```
acostarelli: We should use the jump_utils.jl where possible.

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1635 @@
...
+        meta = service_name,
+    )
+    reserve_deployment_variable = get_variable(container, U, R, service_name)
+    mult_default = get_variable_multiplier(U, R, F)
+    for outage in associated_outages
+        associated_devices =
+            PSY.get_associated_components(sys, outage; component_type = PSY.Generator)
+        outage_id = string(IS.get_uuid(outage))
+        for device in contributing_devices
+            mult = device in associated_devices ? 0.0 : mult_default
```
acostarelli: I think if we're going to be multiplying be 0, we shouldn't add to the expression at all.

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1635 @@
+# ----------------------------------------------------------------------------
```
acostarelli, line 1: Can you explain to me the flow through this file?

src/operation/template_validation.jl
```diff
@@ -294,6 +295,116 @@ function _sc_branch_models(template::IOM.AbstractProblemTemplate)
...
+
+# SC reserve service models in the template.
+function _sc_reserve_service_models(template::IOM.AbstractProblemTemplate)
+    return ServiceModel[
+        m for m in values(get_service_models(template)) if
+        get_formulation(m) <: AbstractSecurityConstrainedReservesFormulation
+    ]
+end
+
+# Whether SC service model `m` should skip `outage`. Dispatched on the outage
```
acostarelli: Unnecessary comment

src/services_models/reserves.jl
```diff
@@ -68,6 +68,15 @@ function get_default_time_series_names(
     )
 end
 
+function get_default_time_series_names(
+    ::Type{<:PSY.Reserve},
+    ::Type{<:AbstractSecurityConstrainedReservesFormulation},
+)
+    return Dict{Type{<:TimeSeriesParameter}, String}(
+        RequirementTimeSeriesParameter => "requirement",
```
acostarelli, line 79: How does adding time series work? Does the user have to use the correct name exactly, add to the system, and hope the component finds it?

src/ac_transmission_models/security_constrained_branch.jl
```diff
@@ -267,47 +267,38 @@ function _has_other_v_container(
...
-`(outage_id::String, monitored_name::String, t::Int)` holding `JuMP.AffExpr`
-zeros for every entry produced by `_resolve_monitored_arcs`. The pre-fill is
-required so the parallel PTDF expression build below cannot race on Dict resize.
+Register an empty `SparseAxisArray` keyed by
+`(outage_id::String, monitored_name::String, t::Int)`. The `String[], String[],
+time_steps` axes only fix the key tuple type; the populated `(outage, name, t)`
+set is ragged, so the build loop fills entries by assignment. Storage stays
+sparse — only assigned keys exist, never the full outage×component×time product.
 """
 function _add_post_contingency_sparse_expression!(
```
acostarelli: We don't need this helper anymore, it's only one line.

src/ac_transmission_models/security_constrained_branch.jl
```diff
@@ -267,47 +267,38 @@ function _has_other_v_container(
...
+    )
 end
 
 """
 Register an empty `SparseAxisArray` keyed by
 `(outage_id::String, monitored_name::String, t::Int)` for the given constraint
-type and meta tag.
+type and meta tag. The build loop fills entries by assignment.
 """
 function _add_post_contingency_sparse_constraints!(
```
acostarelli: We also don't need this helper anymore.

src/ac_transmission_models/security_constrained_branch.jl
```diff
@@ -412,15 +403,34 @@ function add_constraints!(
...
-    # Local relaxation-slack containers keyed by `(outage_id, name, t)`. Built
-    # here (not via `add_variables!`) because the post-contingency axes are only
-    # known after `_resolve_monitored_arcs`; registered after the loop iff used.
-    slack_ub = SparseAxisArray(Dict{Tuple{String, String, Int}, JuMP.VariableRef}())
-    slack_lb = SparseAxisArray(Dict{Tuple{String, String, Int}, JuMP.VariableRef}())
+    # Relaxation-slack containers keyed by `(outage_id, name, t)`, registered only
+    # when slacks are enabled. Built here (not via `add_variables!`) because the
+    # post-contingency keys are ragged; the axes only fix the key tuple type and
+    # the build loop assigns the resolved entries.
+    slack_ub =
```
acostarelli: Can we combine these two if statements

src/ac_transmission_models/security_constrained_branch.jl
```diff
@@ -412,15 +403,34 @@ function add_constraints!(
 
     resolved = _resolve_monitored_arcs(device_model, net_reduction_data)
 
-    con_lb = _add_post_contingency_sparse_constraints!(container, T, V; meta = "lb")
-    con_ub = _add_post_contingency_sparse_constraints!(container, T, V; meta = "ub")
+    con_lb =
+        _add_post_contingency_sparse_constraints!(container, T, V, time_steps; meta = "lb")
```
acostarelli: This meta string is used later to find shared post-contingency constraint sources, so we could put this meta in a constant so we know we're referencing the same string.

src/core/constraints.jl
```diff
@@ -454,6 +494,32 @@ S_k^{\\max}`` for ``k \\in \\{f, t\\}`` via one of two shapes, selected by the
 struct HVDCVSCApparentPowerLimitConstraint <: ConstraintType end
 
 abstract type PowerVariableLimitsConstraint <: ConstraintType end
+
+abstract type PostContingencyVariableLimitsConstraint <: PowerVariableLimitsConstraint end
+
+"""
+Struct to create the constraint to limit post-contingency active power expressions.
```
acostarelli: Why are we suddenly saying "struct to create the constraint" -- are these constraint structs used differently than the others? This struct is the constraint. We can just start with "Constraint to limit..."

src/core/expressions.jl
```diff
@@ -6,7 +6,11 @@ struct ComponentReserveUpBalanceExpression <: ExpressionType end
...
 struct PostContingencyNodalActivePowerDeployment <: PostContingencyExpressions end
+struct PostContingencyAreaActivePowerDeployment <: PostContingencyExpressions end
```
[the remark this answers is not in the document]
acostarelli: This is defined in standard_variables_expressions.jl

src/core/variables.jl
```diff
@@ -204,19 +205,43 @@ Docs abbreviation: ``f^\\text{sl,lo}``
 """
 struct FlowActivePowerSlackLowerBound <: AbstractACActivePowerFlow end
 
+"""
+Struct to dispatch the creation of Post-Contingency Active Power Change Variables.
```
acostarelli: None of this "struct to" stuff. These are all structs.

src/core/constraints.jl
acostarelli: We probably don't even need to say "Constrain to" this file is all constraints.

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1584 @@
...
+exist, never the full outage×component×time product.
+"""
+function _add_service_post_contingency_sparse_expression!(
+    container::OptimizationContainer,
+    ::Type{T},
+    ::Type{R},
+    service_name::String,
+    time_steps::UnitRange{Int},
+) where {T <: PostContingencyExpressions, R <: PSY.AbstractReserve}
+    return IOM.add_expression_container!(
```
acostarelli: Helper unnecessary now

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1584 @@
...
+"""
+function _add_service_post_contingency_sparse_constraints!(
+    container::OptimizationContainer,
+    ::Type{T},
+    ::Type{R},
+    service_name::String,
+    time_steps::UnitRange{Int};
+    meta_suffix::String,
+) where {T <: ConstraintType, R <: PSY.AbstractReserve}
+    return IOM.add_constraints_container!(
```
acostarelli: Helper unnecessary now

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1584 @@
...
+        container, T, R, String[], String[], time_steps;
+        sparse = true, meta = "$(service_name)_$(meta_suffix)",
+    )
+end
+
+"""
+Sparse slack variable container keyed by
+`(outage_id::String, monitored_name::String, t::Int)`. Each entry is a
+non-negative `JuMP.VariableRef` whose objective contribution is
+`POST_CONTINGENCY_CONSTRAINT_VIOLATION_SLACK_COST`. The container starts empty —
```
acostarelli: We don't need to specify how the container is filled here.

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1584 @@
...
+    post_cont_flow,
+    slack_ub,
+    slack_lb,
+    outage_id::String,
+    name::String,
+    t::Int,
+    lb,
+    ub,
+)
+    if slack_ub === nothing
```
acostarelli: Do we gain anything by dispatching on Nothing instead of branching?

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1584 @@
...
+                    continue
+                end
+                ub = get_variable_upper_bound(
+                    variable_type,
+                    service,
+                    device,
+                    F,
+                )
+                ub === nothing || JuMP.set_upper_bound(v, ub)
+                lb = get_variable_lower_bound(
```
acostarelli: The bounds and warm start don't change per-time-step.

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1584 @@
...
+    )
+    reserve_deployment_variable = get_variable(container, U, R, service_name)
+    mult_default = get_variable_multiplier(U, R, F)
+    for outage in associated_outages
+        associated_devices =
+            PSY.get_associated_components(sys, outage; component_type = PSY.Generator)
+        outage_id = string(IS.get_uuid(outage))
+        for device in contributing_devices
+            name = PSY.get_name(device)
+            mult = device in associated_devices ? 0.0 : mult_default
```
acostarelli: If we're gonna multiply by 0, we shouldn't add to expression at all. And we should use our jump utils here not the direct add to expression.

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1584 @@
...
+    )
+    reserve_deployment_variable = get_variable(container, U, R, service_name)
+    mult_default = get_variable_multiplier(U, R, F)
+    network_reduction = get_network_reduction(network_model)
+    for outage in associated_outages
+        associated_devices =
+            PSY.get_associated_components(sys, outage; component_type = PSY.Generator)
+        outage_id = string(IS.get_uuid(outage))
+        for device in contributing_devices
+            mult = device in associated_devices ? 0.0 : mult_default
```
acostarelli: Apply what I commented earlier everywhere.

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1584 @@
...
+                    mult,
+                    reserve_deployment_variable[outage_id, name, t],
+                )
+            end
+        end
+    end
+    return
+end
+
+function add_to_expression!(
```
acostarelli, line 527: I see now two add_to_expression! methods per expression type. They are awfully similar the two pairs. Could they be combined with a helper? If not, I need a little docstring to explain the difference between the two because I can't immediately tell based on the type signature.

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1584 @@
...
+    nodal_deployment = get_expression(container, PostContingencyNodalActivePowerDeployment,
+        R,
+        service_name,
+    )
+    ptdf = get_PTDF_matrix(network_model)
+
+    # Cache pre-contingency flow expressions by monitored type and PTDF columns
+    # by arc — multiple outages may monitor the same arc, and `ptdf[arc, :]` is
+    # an expensive KLU solve we don't want to repeat per outage/time.
+    pre_flow_cache = Dict{DataType, Any}()
```
acostarelli, line 728: I don't see what this cache is saving us. How is indexing the PTDF matrix doing a KLU solve?

src/ac_transmission_models/security_constrained_branch.jl
```diff
@@ -412,15 +368,36 @@ function add_constraints!(
...
 
     use_slacks = get_use_slacks(device_model)
-    # Local relaxation-slack containers keyed by `(outage_id, name, t)`. Built
-    # here (not via `add_variables!`) because the post-contingency axes are only
-    # known after `_resolve_monitored_arcs`; registered after the loop iff used.
-    slack_ub = SparseAxisArray(Dict{Tuple{String, String, Int}, JuMP.VariableRef}())
-    slack_lb = SparseAxisArray(Dict{Tuple{String, String, Int}, JuMP.VariableRef}())
+    # Relaxation-slack containers keyed by `(outage_id, name, t)`, registered only
+    # when slacks are enabled. Built here (not via `add_variables!`) because the
+    # post-contingency keys are ragged; the axes only fix the key tuple type and
```
luke-kiernan, line 385: I had to look up what "ragged" means. Not sure if that's common parlance

src/core/formulations.jl
```diff
@@ -332,6 +332,36 @@ struct RampReserve <: AbstractReservesFormulation end
...
+abstract type AbstractSecurityConstrainedReservesFormulation <: AbstractReservesFormulation end
+
+"""
+Security-constrained contingency reserve formulation: deploys reserves under
+each G-1 outage scoped to the reserve `PSY.Service`. The set of contingencies a
+service responds to is the `PSY.Outage` supplemental attributes attached to that
+service via `add_supplemental_attribute!(sys, service, outage)`; template
+validation mirrors those attachments into `service_model.outages`.
+A `RequirementTimeSeriesParameter` is optional: when present the requirement /
+ramp / participation stack is built; when absent, per-generator
```
luke-kiernan, line 345: > the requirement / ramp / participation stack is built

huh?

src/operation/template_validation.jl
luke-kiernan, line 418: Worth adding something about what's `per_type` here imo: looks like it's the set of monitored components of each datatype?

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1642 @@
...
+# ----------------------------------------------------------------------------
+
+"""
+Resolve every monitored component in `service_model.outages` to a container
+name and arc tuple in the active network reduction. Mirrors the device-side
+`_resolve_monitored_arcs` but operates on a `ServiceModel`. Outages whose
+monitored component types are not modeled in the network are skipped.
+
+Returns
+`Vector{Pair{UUID, Vector{Tuple{DataType, String, Tuple{Int,Int}, String}}}}`
```
luke-kiernan: Yikes, that return type. I'd suggest doing a light rewrite so it's more self-documenting: even just doing something like `const MonitoredArcEntry = Tuple{DataType, String, Tuple{Int,Int}, String}}` would help. Could also define getters.

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1642 @@
...
+        container, T, R, String[], String[], time_steps;
+        sparse = true, meta = service_name,
+    )
+    for (uuid, entries) in resolved
+        outage_id = string(uuid)
+        for (_, name, _, _) in entries
+            for t in time_steps
+                v = JuMP.@variable(
+                    jump_model,
+                    base_name = "$(T)_$(R)_$(service_name)_{$(outage_id), $(name), $(t)}",
```
luke-kiernan, line 218: 1.  String manipulations are slooow
2. Hard-coded formatting is neither modular nor maintainable.  If this is going to see much use, write a lightweight interface instead, e.g.
```julia
get_contigency_base_name(...) = "$(T)_$(R)_$(service_name)_{$(outage_id), $(name), $(t)}"
# ...
get_service_name(contigency_base_name::String) = split(contigency_base_name, '_')[3]
```
so that consumers can work with them, independent of the formatting pattern.

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -0,0 +1,1642 @@
...
+        container, PostContingencyFlowActivePowerSlackLowerBound,
+        service, service_name, resolved, F,
+    )
+    return (slack_ub, slack_lb)
+end
+
+# Emit the `[lb, ub]` post-contingency flow inequalities for one
+# `(outage_id, name, t)`. Slacks-disabled and slacks-enabled paths dispatch on the
+# slack arguments rather than branching at run time.
+function _add_post_contingency_flow_rate_constraint!(
```
luke-kiernan, line 259: generic enough that there's probably an IOM helper. also lots of untyped args.

luke-kiernan: Addressed in stacked PR.

src/services_models/static_injection_security_constrained_models.jl
luke-kiernan: Addressed in stacked PR

src/services_models/static_injection_security_constrained_models.jl
luke-kiernan: Never mind. This sort of thing is common elsewhere in the code too.

luke-kiernan (changes requested):
Big-picture structural issues:

1. Type instability in code downstream of `typeof`. Things like the following are bad:
```julia
        # types cannot determined...
        variable = get_variable(container, U, typeof(d))
        mult = get_variable_multiplier(U, typeof(d), F)
        # ...which forces runtime function resolution in what follows
        for t in time_steps
            add_proportional_to_jump_expression!(
                expression[outage_id, t],
                # ambiguous: compiler cannot predict correct get_index
                variable[name, t], 
                mult,
            )
        end
```
Ideally you'd avoid the `typeof` entirely, but doesn't look possible here without major changes. Adding function barrier helps. [I don't fully understand where to put the function barrier and why.] I had Claude code address this: see #166 [stacked on top of #148]
2. Code repetition: I'm trying to set a new precedent for handling copperplate vs nodal vs area. Instead of writing 3 functions, write one and then multiple-dispatch over what to index on. See #167 [stacked on top of #166]


There's a couple places where I'm seeking clarification.

**edit**: oh also tests. I only went thru `src` thoroughly. I'll leave it up to you to check test reasonableness and coverage.

jd-lara:
I'll close this because this effort needs to be restarted to avoid the multiple conflicts

PR #160, opened 2026-06-29
Move EM/DM to IOM

acostarelli:
I also combined E/D Problems into one 

I'm open to disagreement about this. I don't really know what other infrastructure systems need, but this code seems non-specific to power.


src/core/problem_types.jl
```diff
@@ -1,53 +1,30 @@
...
+Operation problems whose build/validate behavior is fully driven by the
+`ProblemTemplate` (the common case). The seam for a future custom-build problem is
+to subtype `AbstractPowerOperationProblem` directly instead of this.
 """
-abstract type AbstractPowerDecisionProblem <: AbstractPowerOperationProblem end
+abstract type GenericPowerOperationProblem <: AbstractPowerOperationProblem end
 
 """
-Supertype for rolling-horizon (emulation) operation problems solved by `EmulationModel`.
+Default concrete template-driven operation problem. The `M` used when a
```
acostarelli, line 26: This comment is confusing.

src/core/problem_types.jl
```diff
@@ -1,53 +1,30 @@
...
 `DecisionModel`/`EmulationModel` parameterized over a subtype of this is a POM
-operation model; dispatch on `DecisionModel{<:AbstractPowerOperationProblem}` selects the
-POM-side implementations of IOM's extension points.
+operation model.
 """
 abstract type AbstractPowerOperationProblem <: IOM.AbstractOptimizationProblem end
 
 """
-Supertype for single-period (decision) operation problems solved by `DecisionModel`.
+Operation problems whose build/validate behavior is fully driven by the
```
acostarelli, line 19: Don't like this docstring

PR #166, opened 2026-07-01
Contain type instability in G-1 outage-driven typeof dispatches

luke-kiernan:
## Summary

Stacked on #148. Reviewing that PR surfaced type-instability from `typeof(d)`-driven variable/multiplier lookups scattered through the outage/generator code — PSY's `get_component_supplemental_attribute_pairs`/`get_contributing_devices` return heterogeneous, abstractly-typed collections, so these dispatches are unavoidable without a much larger refactor (discussed inline, not pursued here as low-leverage for the cost).

- Wrapped each `typeof(d)`-driven lookup's downstream per-`t` loop in a function barrier, so the dynamic dispatch is isolated to once per device/outage instead of leaking into the inner loop (verified with `@code_warntype`: the loop body compiles fully specialized post-barrier).
- Hoisted one lookup (`_outaged_generator_terms`) that was redundantly re-executed per `(entries, t)` pair despite not depending on either, down to once per outage.
- Routed two constraint sites through `IOM.add_range_bound_constraint!`/`add_range_equality_constraint!` where they cleanly fit the existing `(name, t)` signature, matching established convention (`storage_models.jl`, `thermal_generation.jl`). Deliberately did *not* extend IOM's API (e.g. a `Tuple`-index overload) to force-fit the 3-dim outage containers into the same helpers — that would've bent a shared abstraction around a single caller for marginal benefit; those sites stay as direct `JuMP.@constraint` calls.

## Test plan

- [x] `julia --project=test test/runtests.jl test_static_injection_security_constrained_models` — 1219/1219 pass
- [x] `julia --project=test test/runtests.jl test_services_constructor test_device_hvdc test_device_branch_constructors` — 162/162 pass
- [x] Formatter run (`scripts/formatter/formatter_code.jl`)

### DIFF
--- a/src/services_models/reserves.jl
+++ b/src/services_models/reserves.jl
@@ -211,9 +211,9 @@ function add_constraints!(
                 _sum_reserve_variables(@view(reserve_variable[:, t]), extra)
             use_slacks &&
                 JuMP.add_to_expression!(resource_expression, slack_vars[t])
-            IOM.add_range_bound_constraint!(
-                IOM.LowerBound(), jump_model, constraint, service_name, t,
-                resource_expression, ts_vector[t] * requirement,
+            constraint[service_name, t] = JuMP.@constraint(
+                jump_model,
+                resource_expression >= ts_vector[t] * requirement
             )
         end
     end
@@ -266,9 +266,9 @@ function add_constraints!(
                     var_r[name, t] <= (requirement * max_participation_factor) * param[t]
                 )
         else
-            IOM.add_range_bound_constraint!(
-                IOM.UpperBound(), jump_model, cons, name, t, var_r[name, t],
-                (requirement * max_participation_factor) * ts_vector[t],
+            cons[name, t] = JuMP.@constraint(
+                jump_model,
+                var_r[name, t] <= (requirement * max_participation_factor) * ts_vector[t]
             )
         end
     end
@@ -308,10 +308,8 @@ function add_constraints!(
         resource_expression =
             _sum_reserve_variables(@view(reserve_variable[:, t]), extra)
         use_slacks && JuMP.add_to_expression!(resource_expression, slack_vars[t])
-        IOM.add_range_bound_constraint!(
-            IOM.LowerBound(), jump_model, constraint, service_name, t,
-            resource_expression, requirement,
-        )
+        constraint[service_name, t] =
+            JuMP.@constraint(jump_model, resource_expression >= requirement)
     end
 
     return
@@ -413,9 +411,9 @@ function add_constraints!(
         for d in ramp_devices, t in time_steps
             name = PSY.get_name(d)
             ramp_limits = PSY.get_ramp_limits(d, PSY.SU)
-            IOM.add_range_bound_constraint!(
-                IOM.UpperBound(), jump_model, con_up, name, t, variable[name, t],
-                ramp_limits.up * time_frame,
+            con_up[name, t] = JuMP.@constraint(
+                jump_model,
+                variable[name, t] <= ramp_limits.up * time_frame
             )
         end
     else
@@ -452,9 +450,9 @@ function add_constraints!(
         for d in ramp_devices, t in time_steps
             name = PSY.get_name(d)
             ramp_limits = PSY.get_ramp_limits(d, PSY.SU)
-            IOM.add_range_bound_constraint!(
-                IOM.UpperBound(), jump_model, con_down, name, t, variable[name, t],
-                ramp_limits.down * time_frame,
+            con_down[name, t] = JuMP.@constraint(
+                jump_model,
+                variable[name, t] <= ramp_limits.down * time_frame
             )
         end
     else
--- a/src/services_models/reserves.jl
+++ b/src/services_models/reserves.jl
@@ -211,9 +211,9 @@ function add_constraints!(
                 _sum_reserve_variables(@view(reserve_variable[:, t]), extra)
             use_slacks &&
                 JuMP.add_to_expression!(resource_expression, slack_vars[t])
-            constraint[service_name, t] = JuMP.@constraint(
-                jump_model,
-                resource_expression >= ts_vector[t] * requirement
+            IOM.add_range_bound_constraint!(
+                IOM.LowerBound(), jump_model, constraint, service_name, t,
+                resource_expression, ts_vector[t] * requirement,
             )
         end
     end
@@ -266,9 +266,9 @@ function add_constraints!(
                     var_r[name, t] <= (requirement * max_participation_factor) * param[t]
                 )
         else
-            cons[name, t] = JuMP.@constraint(
-                jump_model,
-                var_r[name, t] <= (requirement * max_participation_factor) * ts_vector[t]
+            IOM.add_range_bound_constraint!(
+                IOM.UpperBound(), jump_model, cons, name, t, var_r[name, t],
+                (requirement * max_participation_factor) * ts_vector[t],
             )
         end
     end
@@ -308,8 +308,10 @@ function add_constraints!(
         resource_expression =
             _sum_reserve_variables(@view(reserve_variable[:, t]), extra)
         use_slacks && JuMP.add_to_expression!(resource_expression, slack_vars[t])
-        constraint[service_name, t] =
-            JuMP.@constraint(jump_model, resource_expression >= requirement)
+        IOM.add_range_bound_constraint!(
+            IOM.LowerBound(), jump_model, constraint, service_name, t,
+            resource_expression, requirement,
+        )
     end
 
     return
@@ -411,9 +413,9 @@ function add_constraints!(
         for d in ramp_devices, t in time_steps
             name = PSY.get_name(d)
             ramp_limits = PSY.get_ramp_limits(d, PSY.SU)
-            con_up[name, t] = JuMP.@constraint(
-                jump_model,
-                variable[name, t] <= ramp_limits.up * time_frame
+            IOM.add_range_bound_constraint!(
+                IOM.UpperBound(), jump_model, con_up, name, t, variable[name, t],
+                ramp_limits.up * time_frame,
             )
         end
     else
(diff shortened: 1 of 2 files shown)


src/services_models/reserves.jl
```diff
@@ -211,9 +211,9 @@ function add_constraints!(
                 _sum_reserve_variables(@view(reserve_variable[:, t]), extra)
             use_slacks &&
                 JuMP.add_to_expression!(resource_expression, slack_vars[t])
-            constraint[service_name, t] = JuMP.@constraint(
-                jump_model,
-                resource_expression >= ts_vector[t] * requirement
+            IOM.add_range_bound_constraint!(
+                IOM.LowerBound(), jump_model, constraint, service_name, t,
```
acostarelli: Not an urgent change, but we should probably change this to accept `IOM.LowerBound` as a `Type` to be consistent.

Btw, why did we decide on types versus instances for dispatch? I see instances being more common in other Julia that I read, but I would imagine it could be slightly slower? Unless the compiler does some magic.

acostarelli: Also, why do we have these IOM helpers in the first place? Is it some compiler benefit?

src/services_models/reserves.jl
```diff
@@ -411,9 +413,9 @@ function add_constraints!(
         for d in ramp_devices, t in time_steps
             name = PSY.get_name(d)
             ramp_limits = PSY.get_ramp_limits(d, PSY.SU)
-            con_up[name, t] = JuMP.@constraint(
-                jump_model,
-                variable[name, t] <= ramp_limits.up * time_frame
+            IOM.add_range_bound_constraint!(
```
acostarelli: I think in general we should be preferring to use the `lower_bound` and `upper_bound` properties of JuMP variables as opposed to constraints. And this feels like a case where we could make a `get_variable_lower_bound`/etc. that dispatches on a service instead of a device, if we don't have such a thing already.

src/services_models/reserves.jl
```diff
@@ -491,6 +493,11 @@ function add_constraints!(
     reserve_response_time = PSY.get_time_frame(service)
     jump_model = get_jump_model(container)
     for d in contributing_devices
+        # `contributing_devices` is flattened across every device type the
+        # service applies to, so `typeof(d)` is runtime-only and this
+        # `get_variable` dispatches dynamically. Hand the resulting `varstatus`
+        # to a function barrier so the `t` loop runs fully specialized instead
+        # of paying a dynamic dispatch on every iteration.
```
acostarelli: Makes sense, I wonder if this is an issue across all service formulations.

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -102,28 +102,37 @@ end
...
-where each inner tuple is `(monitored_type, container_name, arc,
-reduction_kind)`. Outages are sorted by UUID for deterministic axes.
+Returns a [`ResolvedMonitoredArcs`](@ref); each [`MonitoredArcEntry`](@ref) is
+`(monitored_type, container_name, arc, reduction_kind)`.
 """
 function _resolve_service_monitored_arcs(
     service_model::ServiceModel,
     net_reduction_data::PNM.NetworkReductionData,
-)
+)::ResolvedMonitoredArcs
```
acostarelli, line 129: Is this here just for clarity, or does it help performance? I could be in favor of either reason, but if only the former, it does stick out to me as unusual.

src/services_models/static_injection_security_constrained_models.jl
acostarelli, line 532: Do we also want to create an alias for this like you did for `MonitoredArc`s?

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -471,15 +476,33 @@ function add_to_expression!(
         outage in associated_outages || continue
         outage_id = string(IS.get_uuid(outage))
         name = PSY.get_name(d)
+        # `d`'s concrete type is only known at runtime (heterogeneous
+        # outage-generator map), so `get_variable` dispatches dynamically here.
+        # The lookup below is a function barrier: it isolates that one dynamic
+        # dispatch, letting the per-`t` loop inside run fully type-specialized.
```
acostarelli: RE: my comment before

This comment is a bit bulky to put in the middle of the code and twice now. I would prefer it as a shorter comment above the helper definition.

Or maybe we should be handling this better for services in general. Like if every service has this type issue, maybe model constructors should get built per-contributing-device-type.

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -471,15 +476,33 @@ function add_to_expression!(
...
-            )
-        end
+        _add_pre_contingency_terms_over_time!(
+            expression, variable, mult, outage_id, name, time_steps,
+        )
+    end
+    return
+end
+
+function _add_pre_contingency_terms_over_time!(
```
acostarelli, line 565: This is super general, this helper. Maybe due for a refactor one day.

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -578,16 +601,33 @@ function add_to_expression!(
...
-            )
-        end
+        _add_pre_contingency_nodal_terms_over_time!(
+            expression, variable, mult, outage_id, bus_number, name, time_steps,
+        )
+    end
+    return
+end
+
+function _add_pre_contingency_nodal_terms_over_time!(
```
acostarelli: Yeah this helper is quite like the previous one just with one more expression axis. I wonder if we could make a generated function to handle these and put it somewhere in IOM.

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -680,16 +720,33 @@ function add_to_expression!(
...
-            )
-        end
+        _add_pre_contingency_area_terms_over_time!(
+            expression, variable, mult, outage_id, area_name, name, time_steps,
+        )
+    end
+    return
+end
+
+function _add_pre_contingency_area_terms_over_time!(
```
acostarelli: This is identical to the previous helper.

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -680,16 +720,33 @@ function add_to_expression!(
         outage in associated_outages || continue
         outage_id = string(IS.get_uuid(outage))
         name = PSY.get_name(device)
+        # See the function-barrier note in the analogous per-outage power-balance
```
acostarelli: Again these comments here are bulky.

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -1075,31 +1169,58 @@ function add_post_contingency_flow_expressions!(
...
+                )
                 expression_container[outage_id, name, t] = expr
             end
         end
     end
     return expression_container
 end
 
+"""
+Per-outage, once: resolve each outaged generator's area name and its
```
acostarelli, line 1009: I'm thinking we find one place to explain this issue of unknown device types and leave everything else uncommented.

acostarelli (changes requested):
Main takeaways:
- Some redundancies in helpers and comments
- If this type issue is everywhere in services, maybe we should think of a more robust solution

src/services_models/static_injection_security_constrained_models.jl
luke-kiernan: Just for clarity and readability. No performance benefit.

src/services_models/reserves.jl
luke-kiernan: > Also, why do we have these IOM helpers in the first place? Is it some compiler benefit?

PSI is full of code blocks that are basically "copy-paste then change 2-3 small things." In my initial PSI-into-IOM-plus-POM refactor, I scoped out that repetition, then crafted helpers to combine. So: reduce code repetition and increase readability.

Type vs instance: for hard-coded values, should be no perf difference. But type would be the more principled choice, yes.

src/services_models/static_injection_security_constrained_models.jl
luke-kiernan: agreed. I'll shorten it up

src/services_models/reserves.jl
luke-kiernan: Yeah I could get on board with that. We do something similar for ramp limits in IOM. It's just the question of how much do we bite off in this one PR, and how much to differ to later

src/services_models/reserves.jl
acostarelli: Sorry I don't understand, this is a performance benefit?

In this case, I'm not sure that this increases readability for me. My instinctive first thought is "what's a range bound?" . And it's not much code to deduplicate -- it's replacing 2 lines with 2 very similar looking lines, IMO.

But I'm not familiar with all the other call sites so maybe I'm understating the benefit.

src/services_models/static_injection_security_constrained_models.jl
luke-kiernan: Yeah...this function is purely introduced for type stability reasons. Maybe I could make it an anonymous function so that's clearer...

src/services_models/reserves.jl
acostarelli: We have an issue open to make sure variables have bounds somewhere, I think we can try to cover that at the same time. I'll make the link.

src/services_models/static_injection_security_constrained_models.jl
luke-kiernan: Same as above: introduced purely for type stability triage. Also I might've addressed this a bit in #167

src/services_models/static_injection_security_constrained_models.jl
luke-kiernan: agreed, will cut down the other comments to 1-2 lines, `# PERF: function barrier for type stability reasons`

src/services_models/reserves.jl
luke-kiernan: Sure, I'll revert it. I don't want to add layers of indirection merely for indirection's sake

Looking at call sites in Storage, there it's a paired upper/lower bound. So there the readability win is more substantial.

src/services_models/static_injection_security_constrained_models.jl
luke-kiernan: I didn't but we could, I'm not against it

PR #167, opened 2026-07-01
Consolidate post-contingency balance expressions via target-index dispatch

luke-kiernan:
## Summary

Stacked on #166 (which is itself stacked on #148 / `ac/g1-port`). Reviewing the outage/security-constrained code alongside the existing `_system_expression_type`/`_balance_expression_targets` framework in `add_to_expression.jl` surfaced a natural consolidation: the system-wide, PTDF-nodal, and AreaBalance-area post-contingency balance/deployment expressions differed only in which extra index component sits between `outage_id` and `t` — the surrounding accumulation logic (healthy-device reserve deployment / outaged-generator pre-contingency output) was duplicated three times.

- Introduced `_post_contingency_target_axes`/`_post_contingency_target_index`, mirroring the existing `_balance_expression_targets` pattern: plain multiple dispatch (no closures/higher-order functions), returning a `Tuple` of 0 or 1 extra index components.
- Dispatch is keyed on the expression type `T`, not on `network_model`'s runtime type — `PostContingencyActivePowerBalance` is built unconditionally under every network model (the outer system-wide check), so it must always be axis-less regardless of what `network_model` happens to be at that call site. Only `PostContingencyNodalActivePowerDeployment`/`PostContingencyAreaActivePowerDeployment` are network-model-specific, and their callers already guarantee the matching network model before invoking them. (Caught this the hard way — an earlier version dispatched on `network_model` directly and broke the PTDF path; fixed and re-verified.)
- Six near-duplicate `add_to_expression!` methods collapse into two generic ones + the two small dispatch helpers; the three `_add_pre_contingency_*_terms_over_time!` copies collapse into one.
- As a side effect of unifying the signature, the AreaBalance pre-contingency call site now threads `attribute_device_map` through instead of recomputing it via a redundant `PSY.get_component_supplemental_attribute_pairs` call.

Deliberately did **not** extend this to `PostContingencyGenerationBalanceConstraint`/`PostContingencyCopperPlateBalanceConstraint` — on inspection those aren't just an index-shape difference (the AreaBalance one sums in an extra baseline `ActivePowerBalance` expression the system-wide one doesn't need), so forcing them into the same abstraction would conflate two different concerns for two call sites.

## Test plan

- [x] `julia --project=test test/runtests.jl test_static_injection_security_constrained_models` — 1219/1219 pass
- [x] `Test.detect_ambiguities(PowerOperationsModels)` — 0 ambiguities
- [x] Formatter run

### DIFF
--- a/src/services_models/static_injection_security_constrained_models.jl
+++ b/src/services_models/static_injection_security_constrained_models.jl
@@ -419,8 +419,7 @@ _post_contingency_target_axes(
     ::Type{PostContingencyActivePowerBalance},
     ::NetworkModel,
     ::PSY.System,
-) =
-    ()
+) = ()
 _post_contingency_target_axes(
     ::Type{PostContingencyNodalActivePowerDeployment},
     network_model::NetworkModel,
--- a/src/services_models/static_injection_security_constrained_models.jl
+++ b/src/services_models/static_injection_security_constrained_models.jl
@@ -387,15 +387,88 @@ end
 # Post-contingency power-balance, nodal-deployment, area-deployment
 # expressions. Reserve-deployment contributions and generator-outage
 # contributions are added in separate dispatches to avoid `isa` checks.
+#
+# The three expression types below (system-wide balance, PTDF nodal
+# deployment, AreaBalance area deployment) differ only in the extra index
+# component between `outage_id` and `t` — resolved per *expression type*
+# `T` by `_post_contingency_target_axes`/`_post_contingency_target_index`
+# — so one generic method per dispatch pair covers all three instead of
+# one copy each.
+#
+# Dispatch is keyed on `T`, not on `network_model`'s runtime type:
+# `PostContingencyActivePowerBalance` is built unconditionally under every
+# network model (the outer system-wide check, from
+# `_construct_service_post_contingency_balance!`), so it must always be
+# axis-less regardless of what `network_model` happens to be. Only
+# `PostContingencyNodalActivePowerDeployment`/`PostContingencyAreaActivePowerDeployment`
+# are network-model-specific, and their callers (`_add_post_contingency_network_terms!`)
+# already guarantee the matching `network_model` type before invoking them.
 # ----------------------------------------------------------------------------
 
 """
-Add the healthy generators' reserve **deployments** to the per-outage system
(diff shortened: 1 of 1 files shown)


src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -387,15 +387,88 @@ end
...
+for the area path (`PostContingencyAreaActivePowerDeployment`). Mirrors
+[`_post_contingency_target_index`](@ref), which resolves the same axis for
+a single device.
+"""
+_post_contingency_target_axes(
+    ::Type{PostContingencyActivePowerBalance},
+    ::NetworkModel,
+    ::PSY.System,
+) =
+    ()
```
acostarelli: Strange formatting here

src/services_models/static_injection_security_constrained_models.jl
```diff
@@ -419,6 +492,7 @@ function add_to_expression!(
     expression = lazy_container_addition!(container, T,
         R,
         string.(IS.get_uuid.(associated_outages)),
+        _post_contingency_target_axes(T, network_model, sys)...,
```
acostarelli, line 494: I forget the performance implications here, is using the spread operator bad?

acostarelli (changes requested):
This uses the splat operator a few times, is this a performance concern?

Secondarily, I would really like to cut back on the level of AI comments in this repository. The comments here feel very verbose and seem to assume the reader has the context of what the code was before the change.

Side-thought, my gut feeling is this particular idea for a refactor feels like something that could quickly be over-applied across POM. I would prefer something that feels more "sturdy" -- I'm not sure what I mean by that, but I think it has to do with the splatting.

src/services_models/static_injection_security_constrained_models.jl
luke-kiernan: The refactor here is just to reduce code duplication: the types _should_ be constant-propagated such that `_post_contingency_target_axes` and the `...` are both resolved at compile time. I have not verified that is the case though

src/services_models/static_injection_security_constrained_models.jl
luke-kiernan: fixed

PR #162, opened 2026-06-30
Consolidate RunOfRiver; fix hydro budget interval

acostarelli:
Close #45

PR #176, opened 2026-07-11
Fix hydro market-bid cost support: PWL breakpoint scaling and commitment objective path

rodrigomha:
## Summary

Two fixes that make hydro devices work with `MarketBidCost` / `MarketBidTimeSeriesCost`, plus a stale-trait cleanup:

1. **Market-bid PWL breakpoints were scaled by `max_active_power`** for `HydroGen` only. Breakpoints are already in system units - `PSY.Device`'s default multiplier and the explicit `RenewableGen`/`ElectricLoad`/`Source` overrides are all `1.0` - so every hydro offer curve shrank by the unit's own rating factor (4-45x on realistic data). With a nonzero static min, the shrunken curve top can fall below the dispatch lower bound and the whole model dies at presolve with variable-bound conflicts.
2. **`HydroCommitmentRunOfRiver` + market bid costs errored in the objective path** (the constructor carried a `# this is erroring when there's a market bid cost.` comment). Root cause: IOM's default `add_proportional_cost!` calls the static 4-arg `proportional_cost`, which forwards to `PSY.get_fixed` - undefined for market bid costs. Fixed by routing the `OnVariable` proportional cost through `add_proportional_cost_maybe_time_variant!`, exactly like `ThermalGen` and `ControllableLoad`; static `HydroGenerationCost` keeps its fixed-cost behavior through the 5-arg `proportional_cost`.
3. **`requires_initialization(HydroCommitmentRunOfRiver) = false`**, mirroring `ThermalBasicUnitCommitment`: the formulation has no min up/down-time or ramp constraints and its constructors never consume initial conditions, so inheriting `true` from `AbstractHydroUnitCommitment` forced a pointless initialization sub-model.

With these fixes a `HydroDispatch` unit carrying an hourly market-bid offer curve builds and solves under `HydroCommitmentRunOfRiver` with the same semantics as thermal UC: `p >= limits.min * On` plus the On-gated PWL first-breakpoint offset, so a unit with a nonzero min can commit OFF in hours where its offer is absent.

## Test plan

- New regression testset in `test/test_device_hydro_constructors.jl`: breakpoint multipliers pinned at `1.0` (with a nonzero-limits demo device so a revert fails), commitment-path traits pinned (On-gated PWL min offset; lb), and the dispatch formulations' lower bound at `limits.min` pinned as intended behavior.
- `julia --project=test test/runtests.jl test_device_hydro_constructors`: 351/351 pass.


src/static_injector_models/hydro_generation.jl
jd-lara, line 1: Do we really need a separate method here?

test/test_device_hydro_constructors.jl
```diff
@@ -1151,3 +1151,106 @@ end
...
+    # Lower-bound checks need a NONZERO static min: the demo device's default limits are
+    # (min = 0, max = 0), under which every path returns 0.0. SU units require system
+    # attachment, so attach the demo to a throwaway System first.
+    lb_sys = PSY.System(100.0)
+    lb_bus = PSY.ACBus(; number = 1, name = "LB_B1", available = true,
+        bustype = PSY.ACBusTypes.REF, angle = 0.0, magnitude = 1.0,
+        voltage_limits = (min = 0.9, max = 1.1), base_voltage = 345.0)
+    PSY.add_component!(lb_sys, lb_bus)
+    demo_hydro.bus = lb_bus
+    PSY.add_component!(lb_sys, demo_hydro)
```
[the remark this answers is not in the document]
jd-lara: This is a good suggestion

src/static_injector_models/hydro_generation.jl
rodrigomha: I think so, the default we have does not have that for MBC. Let me check if we can reuse the ThermalGen with a Union to HydroGen

rodrigomha: Following up on this: I looked at consolidating the `OnVariable` `add_proportional_cost!` forwarders. Thermal (`thermal_generation.jl`), the new hydro one here, and the load one (`electric_loads.jl`) are all one-line forwards to `add_proportional_cost_maybe_time_variant!`. Options:

**Option A: single 2-way Union method in `common_models/objective_function.jl`** (my preference)

```julia
add_proportional_cost!(
    ::Type{U},
    devices::IS.FlattenIteratorWrapper{T},
    ::Type{V},
) where {
    U <: OnVariable,
    T <: Union{PSY.ThermalGen, PSY.HydroGen},
    V <: Union{AbstractThermalFormulation, AbstractHydroFormulation},
} = add_proportional_cost_maybe_time_variant!(container, U, devices, V)
```

- Deletes the thermal and hydro forwarders. `common_models/objective_function.jl` is included before both device files, so all Union constituents resolve there.
- Prototyped and verified: compiles clean, `Test.detect_ambiguities` returns 0, `which()` shows thermal and hydro UC pairs hit the Union method while loads with non-interruption formulations still fall to the IOM static default, and the hydro + thermal constructor test files pass (5588/5588).
- Caveat: the signature admits cross pairs like `(ThermalGen, AbstractHydroFormulation)` that never occur in practice. All call sites pass matched pairs from the same `DeviceModel{T, U}`, so it is unreachable rather than a behavior change, but the signature is looser than two separate methods.
- I would keep the ControllableLoad forwarder separate: it pins `V` to concrete `PowerLoadInterruption` on purpose, so other load formulations keep falling through to IOM's static default. A 3-way union risks loosening that for no gain.

**Option B: trait-based opt-in**

A `uses_time_variant_proportional_cost(::Type{T}, ::Type{V}) = false` default plus `true` methods for the thermal/hydro/load-interruption pairs, with one generic `OnVariable` `add_proportional_cost!` branching on the trait. Cleaner domain statement and extensible (a new device family adds one trait method instead of editing a Union), but the fallback branch needs an `invoke` to reach the IOM default, since the generic `OnVariable` method would shadow it. That feels like heavier machinery than two device families justify right now.

**Option C: keep the separate per-device forwarders (current PR state)**

Most explicit dispatch, but three identical bodies, which is what prompted this thread.

In all options the device-specific pieces stay where they are (`skip_proportional_cost`, the static `HydroGenerationCost` `proportional_cost`, `is_time_variant_proportional`); only the thin forwarder moves.

I'd go with A unless folks prefer the trait. Thoughts?

---

**Update:** Option A is implemented and verified locally (will push after review). Full test suite passes: 22202/22202, zero method ambiguities, and dispatch verified - thermal and hydro UC hit the shared Union method while loads with non-interruption formulations still fall through to the IOM static default.

Two test changes rode along:

1. Removed the "Initialization with HydroCommitmentRunOfRiver" testset from `test_initialization_problem.jl`. It asserted the initialization sub-model runs with that formulation in the template, but it only ever passed because `HydroCommitmentRunOfRiver` inherited `requires_initialization = true` and forced the init model for every device. This PR sets it to `false` because the formulation never consumes initial conditions: its constructors add no commitment, ramp, or duration constraints (the IC-consuming constraint types) and never touch the initial-conditions machinery, mirroring `ThermalBasicUnitCommitment`. The other hydro commitment formulations keep their inherited `true`.
2. Fixed a leftover `CopperPlatePowerModel` reference in `test_device_hydro_constructors.jl` (a #162/#177 crossing - the testset landed written against the pre-rename formulation name), now `CopperPlateNetworkModel`.

jd-lara: Option A is fine I think, I wonder if that method should just set it for any generation

jd-lara (approved):

PR #177, opened 2026-07-13
native models cleanupv1

jd-lara:
This is the first pass at the clean up after merging the native network models. We still have some crazy union types, but I need to review the abstractions of the network models that we are using for a cleaner implementation. 

@luke-kiernan this PR also addresses some of your original comments. 

I didn't address the additional clean up of the complex/float operations or the tap control since we should implement those once @m-bossart is done with the transformer winding re-write


luke-kiernan (approved):
Looks good to me.

Aside: I'd still like to find (or collectively decide upon) an answer for the below question, from my comments on the "native models" PR. Out-of-scope here so I'll open an issue for discussion

> Here, we change the meta instead of introducing multiple keys. Why? What's the guiding principle on when to do one versus the other?

PR #179, opened 2026-07-13
native models cleanup v2

jd-lara:
This is the follow up to the clean up, it has a big focus on the documentation too


docs/src/reference/formulation_library.md
```diff
@@ -0,0 +1,553 @@
+# Formulation Library
+
+## [Introduction](@id formulation_intro)
+
+A `PowerOperationsModels` problem is assembled from three independent choices:
+
+  - a **network formulation** (`NetworkModel{N}`) — how the transmission system is represented,
+  - a **device formulation** (`DeviceModel{D, F}`) — how each component type is modelled,
```
luke-kiernan, line 8: > three independent choices:
> ...
> - a **device formulation** (`DeviceModel{D, F}`) — how each component type is modelled,

Above snippets imo are a bit misleading: device formulations are different in that they're made _per device type_, not _per simulation_.

docs/src/reference/formulation_library.md
```diff
@@ -0,0 +1,553 @@
...
+`AreaBalanceNetworkModel` are *not* DCP subtypes.
+
+### What each network model adds
+
+The balance expression containers themselves are allocated up front by
+`initialize_system_expressions!`, before any `construct_*!` runs. A network model does not
+create them; it wires the system slacks into them (argument stage) and closes them with a
+balance constraint (model stage).
+
+| Network model             | Argument stage: variables                                             | Model stage: constraints                                                                                                     |
```
luke-kiernan, line 73: Table is information overload imo. It's reference material (not explanation) diataxis-wise so a concise "info dump" is fine, but...really leaves something to be desired.

Many here are "almost the same" in various ways: `PTDF` and `AreaPTDF`, `ACP` and `ACR`, etc. Graphic design wise, is there a different layout that would visually emphasize those differences and collapse some of the similarities?

It'd also help explicitly write out the shorthands somewhere: `"AC polar"` and `"AC rectangular"` spell out a clear narrative as to what's different, whereas the 1-letter difference between `ACPNetworkModel` and `ACRNetworkModel` is easily lost.

luke-kiernan (commented):
Started reviewing, but have other things I need to get to today

PR #180, opened 2026-07-13
address issue #178

luke-kiernan:
Fix one or two bugs where we're indexing 1D containers with 2D axes and document the places where we're using non-2D containers.


src/energy_storage_models/storage_models.jl
```diff
@@ -1348,7 +1347,7 @@ function add_cycling_charge_without_reserves!(
             PSY.get_conversion_factor(d)
         cycle_count = PSY.get_cycle_limits(d)
         efficiency = PSY.get_efficiency(d)
-        constraint[name, time_steps[end]] = JuMP.@constraint(
+        constraint[name] = JuMP.@constraint(
```
jd-lara: this is not correct, the constraint containers should be matrices even it they are 1 Dimensional. If you do this we go into a  Vector instead of a `Matrix{N, 1}`

jd-lara (commented):
This change break the convention to store Nx1 matrices for Single dimension quantities

src/energy_storage_models/storage_models.jl
luke-kiernan: > this is not correct, the constraint containers should be matrices even it they are 1 Dimensional.

Wait, I thought in Slack you agreed that 1D containers were fine.

As-is, this is a latent bug. The container is 1D, yet we're indexing it by 2D, so it throws a `KeyError`.

I had AI do an inventory of both "actually 1D" (`Vector`) containers and "practically 1D" containers (`Matrix{N,1}`). Take a look at the list. So...all the ones that are constraints of the "actually 1D" type should instead get changed to be "practically 1D?"

jd-lara (approved):

PR #181, opened 2026-07-13
Fix thermal min-up/down rounding and improve parallel-branch interface error messages

acostarelli:
Two small fixes ported from PSI:

- **Thermal min-up/down rounding** — drop the stray `Float64` positional argument in `round(..., RoundUp)` in `_get_data_for_tdc` (`thermal_generation.jl`), which is incompatible with the `RoundingMode` argument.
- **Parallel-branch interface error message** — when a transmission interface names only part of a reduced double-circuit / degree-two chain, list the offending branch names in the `ArgumentError` (re-ported onto `main`'s `_error_msg(T)` helper).

### Ported from PSI
- Sienna-Platform/PowerSimulations.jl#1527 (thermal rounding)
- Sienna-Platform/PowerSimulations.jl#1587 (parallel-branch error message)

The HVDC area-balance fix (#1519) originally bundled with this work is already present on `main`, so it is not included here.

---
Part of splitting the former #154 into focused PRs.

### DIFF
--- a/src/common_models/add_to_expression.jl
+++ b/src/common_models/add_to_expression.jl
@@ -2331,7 +2331,8 @@ function _reduced_entry_in_interface(
     ]
 
     if !allequal(in_interface)
-        throw(ArgumentError(_error_msg(T)))
+        branch_names = [PSY.get_name(x) for x in reduction_entry]
+        throw(ArgumentError(_error_msg(T) * " Offending branches: $(branch_names)."))
     end
     return first(in_interface)
 end


src/static_injector_models/thermal_generation.jl
```diff
@@ -1361,8 +1361,8 @@ function _get_data_for_tdc(
             end
             ini_conds[idx, 1] = ic
             ini_conds[idx, 2] = initial_conditions_off[ix]
-            up_val = round(Float64, time_limits.up * steps_per_hour, RoundUp)
-            down_val = round(Float64, time_limits.down * steps_per_hour, RoundUp)
+            up_val = round(time_limits.up * steps_per_hour, RoundUp)
```
luke-kiernan: `round` picks same type as the argument. If `time_limits.up` and `steps_per_hour` both are ints, then `up_val` will also be an int. Check that one or the other is a float so that `up_val` is consistently a float, not an int

luke-kiernan (commented):
> Thermal min-up/down rounding — drop the stray Float64 positional argument in round(..., RoundUp) in _get_data_for_tdc (thermal_generation.jl), which is incompatible with the RoundingMode argument.

I'm confused: `round(Float64, 1.4, RoundUp)` runs just fine for me in REPL. AI hallucination?

src/static_injector_models/thermal_generation.jl
acostarelli: From PSY docs:
```julia
get_time_limits(
      value::ThermalStandard
  ) -> Union{Nothing, @NamedTuple{up::Float64, down::Float64}}
```
So the argument is definitely a float. The change came from https://github.com/Sienna-Platform/PowerSimulations.jl/pull/1527/ , so it could've been an AI hallucination there but doesn't feel like it to me.

I'm also unsure why this change was made. It definitely wasn't a method error before, and I would be surprised if there was any other issue.

rodrigomha: You need to revert this. Just leave the improving on erroring message.

rodrigomha (changes requested):

src/static_injector_models/thermal_generation.jl
jd-lara: The optimization models take only Floats, it makes no sense to do it with integer

acostarelli: I left a comment below but replying again here. Since the argument we're passing is always a float, the return type will also be a float. Unless we're anticipating that up or down will change types then this should be safe.

PR #183, opened 2026-07-13
Add feedforward arguments for renewable dispatch

acostarelli:
Adds `add_feedforward_arguments!` to both renewable `construct_device!` argument stages and the `get_parameter_multiplier(::VariableValueParameter, ::RenewableGen, ::AbstractRenewableFormulation)` method it requires.

### Ported from PSI
- Sienna-Platform/PowerSimulations.jl#1538

---
Part of splitting the former #154 into focused PRs.


luke-kiernan (approved):
One thing to be aware of: right now, `add_feedforward_arguments!` in POM is just a stub, not actual implementation. While we can merge this, it's kind of strange to do so, because it fails to add any new functionality.

m-bossart (approved):

PR #186, opened 2026-07-13
Support InterconnectingConverter under AreaPTDFNetworkModel

acostarelli:
## What

Adds support for `PSY.InterconnectingConverter` injections under `AreaPTDFNetworkModel`. Previously this combination hit an `error("AreaPTDFPowerModel doesn't support InterconnectingConverter")`; now the converter contributes to the `Area`, AC-bus, and DC-bus `ActivePowerBalance` expressions.

- `src/mt_hvdc_models/HVDCsystems.jl` — replace the error stub with the balance injection.
- `test/test_network_constructors.jl` — one targeted build/solve test.

## Implementation notes (addressing review)

- **AC bus indexed by mapped bus number.** Under `AreaPTDFNetworkModel` the AC-bus balance expression is keyed by *mapped* bus numbers when network reductions are active (`make_system_expressions!`). The method now uses `PNM.get_mapped_bus_number(network_reduction, device_bus)` instead of the raw number, matching the PSI original and avoiding a `KeyError` under reductions.
- **`get_variable_multiplier` for signed contributions**, matching the pattern used by the other converter `add_to_expression!` methods in this file (the DC side keeps the `-1.0` sink sign, as in PSI).

## Test

No PSB system ships with *both* areas and converters, so the test assigns the two 5-bus sides of `sys10_pjm_ac_dc` to two areas (bridged by the existing DC links) and asserts the `AreaPTDFNetworkModel` case **builds and solves** with the `Area`/AC/DC balance expressions and converter variables populated. (This gap is why PSI #1605 shipped the method without a dedicated test — verified there is none upstream.)

## Ported from PSI

- `Sienna-Platform/PowerSimulations.jl#1605` — AreaPTDF `InterconnectingConverter` injection.

## Split note

The broad PSI baseline network-constructor test port that previously rode along in this branch has been moved to **#187** so this PR stays focused on the #1605 feature.

### DIFF
--- a/src/mt_hvdc_models/HVDCsystems.jl
+++ b/src/mt_hvdc_models/HVDCsystems.jl
@@ -319,7 +319,33 @@ function add_to_expression!(
     V <: PSY.InterconnectingConverter,
     W <: AbstractConverterFormulation,
 }
-    error("AreaPTDFNetworkModel doesn't support InterconnectingConverter")
+    variable = get_variable(container, U, V)
+    expression_dc = get_expression(container, T, PSY.DCBus)
+    expression_ac = get_expression(container, T, PSY.ACBus)
+    area_expr = get_expression(container, T, PSY.Area)
+    network_reduction = get_network_reduction(network_model)
+    for d in devices, t in get_time_steps(container)
+        name = PSY.get_name(d)
+        device_bus = PSY.get_bus(d)
+        area_name = PSY.get_name(PSY.get_area(device_bus))
+        bus_number_dc = PSY.get_number(PSY.get_dc_bus(d))
+        bus_number_ac = PNM.get_mapped_bus_number(network_reduction, device_bus)
+        add_proportional_to_jump_expression!(
+            area_expr[area_name, t],
+            variable[name, t],
+            get_variable_multiplier(U, V, W),
+        )
+        add_proportional_to_jump_expression!(
+            expression_ac[bus_number_ac, t],
+            variable[name, t],
+            get_variable_multiplier(U, V, W),
+        )
+        add_proportional_to_jump_expression!(
+            expression_dc[bus_number_dc, t],
+            variable[name, t],
+            -1.0,
+        )
+    end
     return
 end
 
--- a/test/test_network_constructors.jl
+++ b/test/test_network_constructors.jl
@@ -72,3 +72,45 @@ end
         @test "4-5-i_1" in axes(cons)[1]
     end
 end
+
+# Regression test for AreaPTDFNetworkModel + InterconnectingConverter support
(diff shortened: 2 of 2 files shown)


src/mt_hvdc_models/HVDCsystems.jl
```diff
@@ -319,7 +319,32 @@ function add_to_expression!(
...
+        bus_number_ac = PSY.get_number(device_bus)
+        add_proportional_to_jump_expression!(
+            area_expr[area_name, t],
+            variable[name, t],
+            1.0,
+        )
+        add_proportional_to_jump_expression!(
+            expression_ac[bus_number_ac, t],
+            variable[name, t],
+            1.0,
```
acostarelli: The original uses `get_variable_multiplier`, why not here?

src/mt_hvdc_models/HVDCsystems.jl
```diff
@@ -319,7 +319,32 @@ function add_to_expression!(
...
+        )
+        add_proportional_to_jump_expression!(
+            expression_dc[bus_number_dc, t],
+            variable[name, t],
+            -1.0,
+        )
+    end
```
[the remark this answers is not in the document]
acostarelli: Agreed. Is there an existing test in PSI we can use?

test/test_network_constructors.jl
```diff
@@ -55,7 +55,12 @@ end
         ),
     )
     set_device_model!(template, Transformer2W, StaticBranch)
-    ps_model = DecisionModel(template, sys; optimizer = HiGHS_optimizer)
+    ps_model = DecisionModel(
```
acostarelli: Why was this change made?

test/test_network_constructors.jl
```diff
@@ -72,3 +77,309 @@ end
         @test "4-5-i_1" in axes(cons)[1]
     end
 end
+
+# --- Network-constructor coverage ported from PSI test_network_constructors.jl ---
```
acostarelli: Why were all these tests added?

acostarelli (commented):
The original PR adds several `get_parameter_multiplier` methods, why weren't those added here?

test/test_network_constructors.jl
acostarelli: There were network tests that were folded into this PR as part of splitting up #154 . They have now been moved to their own PR.

src/mt_hvdc_models/HVDCsystems.jl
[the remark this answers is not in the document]
acostarelli: Claude reported there was not.

src/mt_hvdc_models/HVDCsystems.jl
acostarelli: Fixed.

PR #188, opened 2026-07-14
native models cleanupv3

jd-lara:
The NetworkModel Abstractions is still weird. That has to be the last step


src/operation/template_validation.jl
```diff
@@ -204,40 +204,63 @@ function _check_security_constrained_three_winding_transformer!(
...
+Trait axis describing which network models a device formulation has a `construct_device!`
+path for. The set of network models a formulation builds under cuts across the formulation
+type hierarchy, so it cannot be expressed as a supertype. Declare one
+[`network_support`](@ref) method per formulation; downstream packages extend the gate the
+same way, which a `Union` alias could not allow.
+"""
+abstract type NetworkSupport end
+"Formulation builds under every network model. Default."
+struct AllNetworks <: NetworkSupport end
+"Formulation needs a full AC network; LPACC's linearized reactive layer cannot build it."
```
rodrigomha: Can we add LPACC's meaning here. So the docstrings for this explain right away what it is.

src/static_injector_models/thermal_generation.jl
```diff
@@ -248,6 +249,18 @@ function get_default_time_series_names(
     )
 end
 
+# FixedOutput has no dispatch decision, so its output is driven entirely by time series
+# rather than by the fuel cost the dispatchable formulations above consume.
+function get_default_time_series_names(
+    ::Type{<:PSY.ThermalGen},
+    ::Type{FixedOutput},
+)
+    return Dict{Any, String}(
```
rodrigomha:
```suggestion
    return Dict{Type{<:TimeSeriesParameter}, String}(
```

I think this is the standard approach we use for the default_time_series_names. See Renewables here: https://github.com/Sienna-Platform/PowerOperationsModels.jl/blob/90bf75b2a765ee795a5b90ab52ba25c094081e47/src/static_injector_models/renewable_generation.jl#L49

rodrigomha (changes requested):
This looks solid. Just minor changes and we can approve.

jd-lara:
Pushed an update — summary of what changed:

**Rebased onto `main`** (picks up #176/#180/#182/#184 and the #179 merge). One conflict in `test/test_device_hydro_constructors.jl` (append-vs-append with the market-bid testset from main); resolved keeping both sides.

**Review feedback addressed** (`8a67144`, `d4691de`):
- @rodrigomha's comments: typed dicts in the ThermalGen `get_default_time_series_names` methods (`Dict{Type{<:TimeSeriesParameter}, String}` for FixedOutput; `Dict{Type{<:ParameterType}, String}` for the FuelCost one, since `FuelCostParameter <: ObjectiveFunctionParameter` and IOM's `DeviceModel` field is keyed on `ParameterType`), and LPACC spelled out in the `AllNetworksExceptLPACC` docstring.
- @luke-kiernan's #179 comments on the formulation library page: intro reworded so the one-per-problem network choice is distinguished from per-device-type/per-service formulations, and a shorthand legend added (ACP/ACR/IVR/LPACC/DCP/DCPLL/NFA/PTDF).
- Items from the posted code review: stale "only Active Power" docstring rescoped onto the two `HydroWaterFactorModel` dispatches, change-narrating comments replaced with invariants, the DCP injection-recovery derivation stated once (in the `_dcp_nodal_injection_expressions` docstring), copy-pasted test comments removed, `_controllable_load_power_factor` hoisted out of the timestep loops, `mock_construct_device!` now delegates to `mock_construct_devices!`, and two testsets hoist their PSB system builds out of the network loops.

**Heads-up: `main` is currently red** (`f809d63` fixes it here). #182 widened `get_default_time_series_names(Source, Union{FixedOutput, AbstractSourceFormulation})` from empty to always naming `max_active_power_out/in`; that flips the `haskey` gates in the ImportExport constructor to always-true, and `add_parameters!` hard-errors on systems without those series — which #179's TS-less `ImportExportSource` testsets exercise. Each PR was green alone; the merged combination fails `test_device_source_constructors` on `main` (reproduced on a clean `origin/main` checkout). Fix: `FixedOutput` keeps the TS defaults (it is time-series driven), `AbstractSourceFormulation` returns an empty typed dict, restoring the opt-in contract (the with-TS testsets already pass explicit `time_series_names`). If preferred, this one-file fix can be cherry-picked to `main` as its own PR.

Verified locally: `test_device_hydro_constructors`, `test_device_load_constructors`, `test_device_thermal_generation_constructors`, `test_ac_transmission_security_constrained_models` (12309 pass), plus `test_device_source_constructors`, `test_import_export_cost`, `test_mbc_parameter_population` (198 pass, including the two testsets that fail on `main`). Full suite run in progress.

src/ac_transmission_models/security_constrained_branch.jl
```diff
@@ -654,6 +663,140 @@ function add_post_contingency_flow_expressions!(
...
+        JuMP.add_to_expression!(injection, -coefficient, variable)
+    end
+    return injection
+end
+
+"""
+Nodal active-power injection expressions on a DCP network, with rows ordered by
+`bus_axis` (the MODF bus axis) so that row `i` multiplies MODF column entry `i`.
+
+The DCP `ActivePowerBalance` cannot be used directly in the MODF product: it
```
luke-kiernan, line 756: Paragraph needs a good edit for clarity. For starters, ambiguous antecedent: "it holds the injections"...meaning the MODF product or the DCP `ActivePowerBalance`?

src/static_injector_models/hydrogeneration_constructor.jl
```diff
@@ -475,19 +475,13 @@ function construct_device!(
     return
```
luke-kiernan, line 475: 2k lines of `construct_device!` calls, yikes.

Questions: 
1. Do we have test coverage for this file and these functions?
2. Have you gone thru and made sure that these are all meaningfully different and can't be easily consolidated? I see you've done some of that already--e.g `_add_hydro_turbine_water_constraints!`.

src/static_injector_models/thermalgeneration_constructor.jl
```diff
@@ -1515,6 +1545,14 @@ function construct_device!(
         device_model,
         network_model,
     )
+    add_to_expression!(
```
luke-kiernan, line 1548: Is this a bugfix, or a change in functionality as to what this formulation means?

src/static_injector_models/electric_loads.jl
```diff
@@ -138,93 +138,103 @@ function add_expressions!(
...
 
+"""
+Electric Load implementation to add the realized shifted load to `ActivePowerBalance`
+expressions for any network model. Targets come from
+[`_balance_expression_targets`](@ref) (nodal bus, system, area, or PTDF/AreaPTDF
+system+bus pair), matching how every other injection enters the balance; the multiplier
+`-1.0` matches `get_variable_multiplier(::Type{<:VariableType}, ::Type{<:PSY.ElectricLoad},
+::Type{<:AbstractLoadFormulation})`, the sign convention shared by every load formulation.
+"""
 function add_to_expression!(
```
luke-kiernan, line 149: The body of this looks highly generic. Do we have a helper for this?

edit: we don't add expressions to expressions elsewhere, so I guess this stays. But still, long term food-for-thought. Among the 10 lines there's only 2 things that are specific to the formulation: the `get_expression` on line 163 and the `-1.0` weight on line 169.

edit^2: separately, super verbose docstring.

src/network_models/instantiate_network_model.jl
```diff
@@ -700,3 +743,82 @@ function _maybe_build_modf_matrix!(
...
     return
 end
+
+# DCP counterpart of `_maybe_build_modf_matrix!`: populate the contingency
+# matrix when the template uses an outage-aware (security-constrained) branch
+# formulation, reconcile its reduction against the Ybus-derived one, and drop
+# outages PNM could not register. Must run before
+# `_finalize_network_reduction!` because reconciliation can replace
+# `model.network_reduction`.
+function _maybe_build_dcp_modf_matrix!(
```
luke-kiernan, line 753: Do we have tests for this?

src/operation/template_validation.jl
```diff
@@ -204,40 +204,66 @@ function _check_security_constrained_three_winding_transformer!(
...
+    network_support(::Type{<:AbstractDeviceFormulation}) -> NetworkSupport
+
+Which network models a device formulation can be constructed under. Defaults to
+[`AllNetworks`](@ref); a formulation whose `construct_device!` is bound to a narrower
+network type must declare it here or it fails deep inside `build!` instead of at template
+validation.
+"""
+network_support(::Type{<:AbstractDeviceFormulation}) = AllNetworks()
+
+# LPACC is reactive-capable at the network level (`network_has_reactive_power` is true), so
```
luke-kiernan, line 246: what?? confusing comment imo

luke-kiernan (changes requested):
Basically: test coverage and comment clarity/conciseness. Otherwise looks good.

src/operation/template_validation.jl
rodrigomha: Yeah this comment is bad.

We should add a comment along the lines:
```julia
# Advanced reactive power control schemes are not supported for
# linear-programming AC cold-start approximation (LPACC) networks

PR #190, opened 2026-07-15
native models cleanupv4

jd-lara:
This is another clean up for the native network models


src/area_interchange.jl
```diff
@@ -234,3 +376,58 @@ function add_constraints!(
...
+    ::Type{U},
+    ::Type{V},
+    names::Vector{String},
+    net_reduction_data::PNM.NetworkReductionData,
+    orientation_sign_cache::Dict{Tuple{DataType, String}, Float64},
+    t::Int,
+) where {
+    U <: Union{FlowActivePowerFromToVariable, FlowActivePowerToFromVariable},
+    V <: PSY.ACBranch,
+}
```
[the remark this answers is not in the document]
rodrigomha: Is this true? Before TwoTerminalHVDC were subtype of ACBranch. Is this not the case anymore?

src/static_injector_models/shunt_constructor.jl
```diff
@@ -147,7 +168,7 @@ function add_constraints!(
             v_arrays[1], bus_name, "connection bus of shunt $(name)",
         )
         for t in time_steps
-            v2 = _bus_voltage_squared(v_arrays, bus_name, t)
+            v2 = _bus_voltage_squared(network_model, v_arrays, bus_name, t)
             cons[name, t] = JuMP.@constraint(jm, q[name, t] == b[name, t] * v2)
```
rodrigomha, line 172: Are we okay with LPACC model stop being linear when we introduce variable shunts? This `b * v2` in LPACC is `b * (1 + 2φ) = b + 2 b φ` that will end up introducing a bilinear term.

This will force for LPACC to use Ipopt, beating the purpose of linear. What if we introduce a warning in the constructor of VariableShunts so the user is warned that adding VariableShunts in LPACC make the model non-linear?

src/area_interchange.jl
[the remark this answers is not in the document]
rodrigomha: Confirmed. This is an hallucination. We should ignore Copilot comment.

src/ac_transmission_models/voltage_control_tap_models.jl
```diff
@@ -214,6 +215,62 @@ function add_constraints!(
...
+            adm.g, adm.b, adm.g_fr, adm.b_fr, adm.g_to, adm.b_to, adm.shift,
+        )
+        for t in time_steps
+            phi_fr = phi[from_bus, t]
+            phi_to = phi[to_bus, t]
+            vad = va[from_bus, t] - va[to_bus, t]
+            cs_sum = cs[name, t] + phi_fr + phi_to
+            dev_fr = 1.0 + 2.0 * phi_fr
+            dev_to = 1.0 + 2.0 * phi_to
+            _add_tap_acr_flow!(
```
rodrigomha: Same here, this call will make `1/tap` and `1/tap * vv_cos` and other surrogates. Making it non-linear. I think we should add a warning in the constructor of this too with LPACC.

rodrigomha (changes requested):
We should update the fix for LCC with AreaInterchange. The warning is optional for LPACC but I think is a good idea.

src/ac_transmission_models/branch_constructor.jl
```diff
@@ -2776,6 +2821,47 @@ function construct_device!(
...
 
+# PNM's reduction maps cover only PSY.ACTransmission, so _get_branch_map never sees the
+# HVDC tie lines; they are also never reduced, so collecting them from the system
+# directly is safe.
+function _add_hvdc_inter_area_branches!(
+    inter_area_branch_map::Dict{Tuple{String, String}, Dict{DataType, Vector{String}}},
+    sys::PSY.System,
+    network_model::NetworkModel,
+)
+    for br_type in network_model.modeled_branch_types
```
luke-kiernan, line 2851: Performance musing: `modeled_branch_types` here is a runtime value. I suspect it'd be faster to loop over concrete subtypes of `TwoTerminalHVDC` [known at compile time] and skip those not in `modeled_branch_types`.

src/ac_transmission_models/voltage_control_tap_models.jl
```diff
@@ -214,6 +215,62 @@ function add_constraints!(
     return
 end
 
+# LPACC (linear-AC) variable-tap Ohm's law. The LPAC substitution replaces the ACR
+# voltage bilinears with their linearized surrogates — vsq_fr → 1 + 2φ_fr,
+# vsq_to → 1 + 2φ_to, vv_cos → cs + φ_fr + φ_to, vv_sin → va_fr - va_to — so the
+# shared `_add_tap_acr_flow!` builder emits constraints that reduce term-for-term to
+# the fixed-tap LPACC StaticBranch form (AC_branches.jl) when `t == tm`.
```
luke-kiernan: This level of details would fit better in the docs. We should put a reference there anyway

src/ac_transmission_models/voltage_control_tap_models.jl
```diff
@@ -490,15 +547,17 @@ function _apply_tap_control_objective!(
...
-# flow objectives are well-defined in current space — identical control logic to ACR.
+# Rectangular (ACR/IVR) and deviation (LPACC): VOLTAGE pins the regulated-bus magnitude
+# via fix_regulated_voltage! — the component-owned (component, "1")
+# RegulatedVoltageMagnitude aux variable under ACR/IVR, the shared bus VoltageDeviation
+# (at setpoint - 1) under LPACC; reactive/active-flow objectives pin the from-to terminal
+# flow. The ACR/IVR aux variable/constraint are added unconditionally in the construction
+# stages, so only the fix is objective-conditional (count-invariance). Under IVR the
+# from-to power variables (pft/qft) are bilinear-linked to the branch currents in the IVR
+# Ohm's law, so the flow objectives are well-defined in current space — identical control
+# logic to ACR.
```
luke-kiernan: Whaaaat...this comment is more confusing than the function itself. Too many superfluous details. What actually matters here? I see:

-  we handle the `RegulatedVoltageMagnitude` (ACR) vs `VoltageDeviation` (LPACC) distinction via multiple dispatching the `fix_regulated_voltage!` call
- from the formulations, we know that `pft` and `qft` are defined

I don't follow this part
> only the fix is objective-conditional (count-invariance)

src/ac_transmission_models/voltage_control_tap_models.jl
luke-kiernan, line 527: Without the `"1"`, what other thing does it collide with? Why `"1"` of all things? (Is there a more descriptive label you could use?)

src/static_injector_models/shunt_constructor.jl
```diff
@@ -27,19 +27,40 @@ function _fetch_voltage_arrays(
...
+    ::NetworkModel{<:Union{ACRNetworkModel, IVRNetworkModel}},
+    vars,
     bus_name::String,
     t::Int,
-) where {V, W}
+)
     return vars[1][bus_name, t]^2 + vars[2][bus_name, t]^2
 end
 
+# LPACC linearization of V² = (1 + φ)² around φ ≈ 0, consistent with the 1 + 2φ
```
luke-kiernan, line 53: Oooh, so _that's_ what `φ` is: `φ := V_mag - 1`. Seems worth stating somewhere, if it isn't already (possible that I just missed it).

src/area_interchange.jl
```diff
@@ -213,17 +213,159 @@ function add_constraints!(
...
+end
+
+# On the AC networks a tie line is lossy, so its two directional active-flow variables
+# differ by the line loss and the interchange must fix a measurement end. It is measured
+# at the exporting (from-area) boundary: each tie line contributes the directional flow
+# variable metered at its terminal inside the interchange's from-area, so the measured
+# export includes the tie-line loss. A branch keyed (from_area, to_area) is metered at
+# its own from terminal (FlowActivePowerFromToVariable); a branch keyed
+# (to_area, from_area) at its own to terminal (FlowActivePowerToFromVariable). Swap the
+# two selections below to measure at the importing boundary instead.
```
luke-kiernan: super verbose comment

src/area_interchange.jl
```diff
@@ -234,3 +376,58 @@ function add_constraints!(
...
+        flow_variable = get_variable(container, U, V)
+        measured_direction_mult = 1.0
+    else
+        # Lossless formulations carry one signed flow variable (positive from -> to):
+        # the measurement end is immaterial and only the interchange direction matters.
+        flow_variable = get_variable(container, FlowActivePowerVariable, V)
+        measured_direction_mult = _measured_direction_mult(U)
+    end
+    for name in names
+        orientation_sign = get!(orientation_sign_cache, (V, name)) do
```
luke-kiernan, line 394: A little redundant to include the HVDCs in an orientation cache when their orientation never changes. I guess it keeps things consistent, though.

src/ac_transmission_models/branch_constructor.jl
luke-kiernan: Wait, is there any network model that _isn't_ covered here? I think this should just be `<: AbstractNetworkModel`.

src/ac_transmission_models/branch_constructor.jl
```diff
@@ -2662,7 +2605,7 @@ function construct_device!(
     network_model::NetworkModel{U},
 ) where {
     T <: Union{StaticBranchUnbounded, StaticBranch},
-    U <: AbstractActivePowerModel,
+    U <: Union{AbstractActivePowerModel, AbstractReactivePowerNetworkModel},
```
luke-kiernan: Same here, looks like this is just `<: AbstractNetworkModel`.

src/ac_transmission_models/security_constrained_branch.jl
luke-kiernan, line 756: Rather verbose. Also I don't follow its reasoning.

src/network_models/instantiate_network_model.jl
luke-kiernan, line 106: verbose comments

src/static_injector_models/hydrogeneration_constructor.jl
luke-kiernan: With the exception of this `add_constraints` call (and the different type on the `network_model` argument), identical to the above.

src/static_injector_models/shunt_constructor.jl
```diff
@@ -224,17 +245,18 @@ function _apply_shunt_control_objective!(
...
+# the bus VoltageDeviation (at setpoint - 1) under LPACC (see
 # fix_regulated_voltage!). All other modes free-optimize (Q adjusts to network
 # needs). The aux variable/constraint are always present under ACR/IVR, so only the
 # fix is mode-conditional (count-invariance).
 function _apply_shunt_control_objective!(
     container::OptimizationContainer,
     devices::IS.FlattenIteratorWrapper{PSY.FACTSControlDevice},
     network_model::NetworkModel{
-        <:Union{ACPNetworkModel, ACRNetworkModel, IVRNetworkModel},
+        <:Union{ACPNetworkModel, ACRNetworkModel, IVRNetworkModel, LPACCNetworkModel},
```
luke-kiernan: This union is precisely `NativeACNetworkModel`.

src/static_injector_models/thermalgeneration_constructor.jl
luke-kiernan, line 49: Wait, you're adding `ReactivePowerBalance`, when the signature says `network_model::NetworkModel{<:AbstractNetworkModel}`. What if the network model doesn't model reactive power?

src/PowerOperationsModels.jl
luke-kiernan: verbose comment.

luke-kiernan (commented):
First pass thru the code. Didn't get to looking at the tests.

src/ac_transmission_models/security_constrained_branch.jl
luke-kiernan: Re-reading it and conversing with Claude, I guess I'd say something like

> We want to recover the bus injections. For DCP, the nodal balance constraint contains `injections[bus] + flow_terms[bus]`, which is identically zero. Hence `injections[bus] == -flow_terms[bus]`. It's easier to isolate the RHS (keep all `FlowActivePowerVariable` terms and negate them) than the LHS (ugly mix of things).

src/static_injector_models/shunt_constructor.jl
rodrigomha: Applied

src/ac_transmission_models/voltage_control_tap_models.jl
rodrigomha: Is the case used for single axis containers with an extra one. I think we can revisit this for later.

src/ac_transmission_models/voltage_control_tap_models.jl
rodrigomha: We reverted the LPACC changes so most of the confusion went away with this.

src/ac_transmission_models/voltage_control_tap_models.jl
rodrigomha: Reverted

src/ac_transmission_models/branch_constructor.jl
rodrigomha: It is, but is super minor, since this runs once per build. There could be some improvements but we need to look at this later with a build profile.

src/static_injector_models/thermalgeneration_constructor.jl
rodrigomha: There is a explicit ActiveNetworkModel so this will only enter to ReactiveNetworkModels.

src/static_injector_models/hydrogeneration_constructor.jl
rodrigomha: Confirmed - the two methods differed only by the reactive-limits call, and it turned out the same duplication existed in seven other hydro formulations. Addressed in 8bb43f0: each ModelConstructStage pair now collapses into a single `AbstractNetworkModel` method via the network-conditional helpers (`_maybe_add_reactive_power_constraints!`, new variable-typed form), taking the file from 41 to 33 constructor methods, with a container-schema parity test pinning that the reactive constraint appears exactly on reactive-capable networks. The ArgumentConstructStage pairs don't collapse cleanly yet - two of them would hit a dispatch ambiguity against a pre-existing overly broad `AbstractHydroDispatchFormulation` argument-stage bound - so those are tracked as a follow-up.

src/area_interchange.jl
rodrigomha: Yeah. Is probably good for consistency here

PR #206, opened 2026-07-23
Per-type service models with merged sparse/dense reserve containers

rodrigomha:
## Summary

Refactors reserve/ancillary-service **service modeling** from the service-name-as-`meta` pattern to **per-type `ServiceModel`s with merged containers**, and builds on that to add **per-device service bids**. One `ServiceModel` now covers every service of a type (like `DeviceModel`); each service fills its own slice of a shared container:

- **Device-indexed reserve entries** (`ActivePowerReserveVariable`, participation/power constraints): one merged **sparse 3D** container keyed `(service_name, device_name, time)`.
- **Requirement-side entries** (`RequirementConstraint`, `ServiceRequirementVariable`, slacks): one **dense 2D** container keyed `(service_name, time)`.

This removes the `aggregated_service_model` / `NO_SERVICE_NAME_PROVIDED` machinery and the per-service-name model API, replacing it with type-keyed `set_service_model!` mirroring `set_device_model!`.

This PR consolidates the full effort (previously stacked as #207 Phase A+B and #210 Phase C; both are now folded in here).

## What's included, by phase

**Core — per-type service models**
- Per-type `ServiceModel` API and per-type construction; sparse service-indexed reserve variables; dense service-indexed requirement containers.
- Removed `aggregated_service_model` / `NO_SERVICE_NAME_PROVIDED`; fixed per-type service-model printing.

**Phase A — TransmissionInterface**
- Merged the interface **slack** and **flow-limit-parameter** containers into one-per-type.

**Phase B — ORDC cost expression**
- Folded the ORDC `ProductionCostExpression` into a merged **dense `(service, time)`** container.

**Phase C1 — ORDC cost parameters**
- Merged the ORDC slope/breakpoint PWL cost `TimeSeriesParameter`s into one container per service type (covariant `Vector{<:ReserveDemandTimeSeriesCurve}` axis + batch-wide tranche padding).
- Retired the now-dead single-service `TimeSeriesParameter` `add_parameters!`/`_add_parameters!` path.

**Phase C2 — per-device service bids (reserve offers)**
- A contributing device can now **OFFER a PWL price/quantity curve** for a reserve (PSY `set_service_bid!` / `get_services_bid`); its reserve award is priced by that curve instead of the flat `DEFAULT_RESERVE_COST`.
- New `PiecewiseLinearBlockReserveOffer` (SparseVariableType, 4D `(service, device, segment, time)`) + `ReserveOfferLinkingConstraint`; `add_reserve_offer_costs!` builds the delta/block-offer formulation (`δ_k ≥ 0`, `δ_k ≤ P_{k+1}-P_k`, `Σ δ_k = award`, `objective += Σ slope_k·δ_k·Δt`). Devices without an offer keep the flat cost via a `skip_devices` pass.
- The 4D container uses IOM's new `sparse_variable_key_type` trait (see #141).

**Cross-cutting**
- **Loud error** when a modeled reserve has no available contributing devices/branches (previously a silent drop), exempting `ConstantReserveGroup`. This unblocks `ConstantReserveGroup` construction, which was previously unbuildable; GroupReserve now has end-to-end test coverage (single/multi-service requirement sums, both reserve directions, requirement enforcement, the unmodeled-member error path, and construction ordering). **Closes #151.**
- **Type-stability quick wins** in reserve construction (function barriers; fixed a latent `UndefVarError` in a `@debug`).
- **`get!` cleanup**: type-as-constructor defaults (`get!(Vector{T}, …)` / `get!(Dict{…}, …)`) instead of anonymous closures in the contributing-devices map builder (per odow's suggestion in #202). The remaining eager `get!` sites in other subsystems are handled separately in #215.

**Docs**
- New **"Running a Problem with Service Bids"** tutorial (`docs/src/tutorials/service_bids.jl`) walking through building a system, attaching per-device reserve offers, solving, and reading results.
- Tutorial pipeline fix: code-block execution now follows the `EXECUTE = TRUE` opt-in (non-executed tutorials render as plain `julia` blocks, so a tutorial needing solvers/data the docs env lacks still builds).

## Dependencies

- **IOM PR #141** (`rh/dev_service_refactor`): the `OptimizationContainer`/store/`ServiceModel` changes and the new `sparse_variable_key_type` trait live in InfrastructureOptimizationModels. `[sources]` here points at that branch; revert to a registered/`main` rev once #141 merges.
- **PowerSystems** pinned to `psy6`, on the current **IS4** InfrastructureSystems tip.

## Testing

Full suite green: **31244 / 31244**, no failures.

The **transformer** control formulations and the **PowerFlows-dependent** suites are currently disabled in `test/runtests.jl` / `test/includes.jl` (`psy6: disabled pending transformer refactor` — `PowerFlows@psy6` does not precompile: `ThreeWindingTransformerWinding` → `ThreeWindingTransformerCircuit`). That gating is inherited from `main` and is why the suite total is lower than pre-rebase; it is not a regression in this PR. Those suites re-enable once the upstream transformer refactor lands.

## Reviewer notes

- Comments tagged `DELETE-AFTER-REVIEW` explain the before/after of the container change; remove once approved (`grep -rn "DELETE-AFTER-REVIEW"`).
- Deferred follow-ups are flagged inline as `TODO(services …)`: the per-(device-type) function barrier over `contributing_devices_map` and the abstract flattened-`Vector` roots (in IOM structs) for type stability; the merged-container full scans in the flat-cost/group-reserve paths (replace with keyed slices); per-service keying pending the reserve-feedforward port.
- `main` is 5 commits ahead (CI/docs only — codecov/actions/Sienna.md); the PR is `MERGEABLE` with no conflicts. A final rebase onto latest `main` is needed at merge time (and will re-enable the transformer/PF suites only once that refactor lands).


test/test_services_constructor.jl
```diff
@@ -127,6 +120,235 @@ end
...
+    empty_reserve = VariableReserve{ReserveUp}("ReserveNoDevices", true, 5.0, 0.1)
+    PSY.add_service!(sys, empty_reserve, PSY.Device[])
+
+    template = get_thermal_dispatch_template_network(CopperPlateNetworkModel)
+    set_service_model!(template, ServiceModel(VariableReserve{ReserveUp}, RangeReserve))
+    @test_throws "ReserveNoDevices" DecisionModel(
+        template,
+        sys;
+        optimizer = HiGHS_optimizer,
+    )
```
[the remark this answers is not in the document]
rodrigomha: This string form is supported

src/services_models/service_slacks.jl
```diff
@@ -1,28 +1,27 @@
-function reserve_slacks!(
+function add_reserve_slacks!(
     container::OptimizationContainer,
-    service::T,
+    ::Type{T},
+    service_names::Vector{String},
 ) where {T <: Union{PSY.Reserve, PSY.ReserveNonSpinning}}
```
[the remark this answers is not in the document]
rodrigomha: We have a union that handles this, so no issue.

docs/src/reference/optimization_container_axes.md
luke-kiernan: > Everything else built via `add_*_container!` is 2D `[device_name, time_step]`

Misleading: should probably be moved to the end of the document. Or add another entry to the "3D containers" chart for the service-indexed service containers

src/common_models/add_expressions.jl
```diff
@@ -207,12 +207,40 @@ function add_expressions!(
...
         PSY.get_name.(devices),
-        time_steps;
-        meta = get_service_name(model),
+        time_steps,
+    )
+    return
+end
+
+"""
+Merged cost-expression container for reserve services (e.g. `ProductionCostExpression` for ORDC):
```
luke-kiernan: The "merged" word here is confusing imo. It only makes sense in the context of the history of the code

src/common_models/add_parameters.jl
```diff
@@ -804,7 +805,7 @@ end
 function _add_parameters!(
```
luke-kiernan, line 806: AI is flagging this as pretty close to the device version, add_parameters.jl:145. Combine? There's some differences--behavior when TS is missing, value unwrapping, resolution handling--so it's not a no-brainer.

src/common_models/add_parameters.jl
```diff
@@ -813,41 +814,44 @@ function _add_parameters!(
...
-            ts_type,
-            service,
-            ts_name;
-            interval = _to_is_interval(ts_interval),
-        ),
-    )
-    @debug "adding" T U _group = IOM.LOG_GROUP_OPTIMIZATION_CONTAINER
-    additional_axes = calc_additional_axes(container, T, [service], model)
+    names = [PSY.get_name(s) for s in services]
+    ts_uuids = [
```
luke-kiernan: Code smell: this is the only place where we use a list of UUIDs as an array axis. Elsewhere UUIDs are solely used for caching.

After some back-and-forth with AI: as-is, you're bulding the uuid list one-per-service with no deduplication, so you might as well use names.

Separately: the device path passes both `interval` _and_ `resolution` to `IS.get_time_series_uuid`. Here we only pass `interval`. Is that purposeful, or a latent bug?

src/core/problem_template.jl
```diff
@@ -243,35 +225,47 @@ function _populate_contributing_devices!(
         empty!(service_models)
         return
     end
+    # One model per service type; fill the per-service nested map for every available
```
luke-kiernan: Comment verbosity alert

src/services_models/reserve_group.jl
```diff
@@ -40,31 +45,49 @@ function add_constraints!(
...
+
     for t in time_steps
         resource_expression = JuMP.GenericAffExpr{Float64, JuMP.VariableRef}()
-        for reserve_variable in reserve_variables
-            JuMP.add_to_expression!(resource_expression, sum(@view reserve_variable[:, t]))
-        end
-        if use_slacks
-            resource_expression += slack_vars[t]
+        for r in contributing_services
+            for var in get(member_vars, (PSY.get_name(r), t), JuMP.VariableRef[])
```
luke-kiernan: perf: per-loop allocations from eager `get`

src/services_models/reserve_offers.jl
```diff
@@ -0,0 +1,92 @@
...
+                nseg = length(slopes)
+                pwl_vars = Vector{JuMP.VariableRef}(undef, nseg)
+                for k in 1:nseg
+                    v =
+                        blk[(service_name, dev_name, k, t)] = JuMP.@variable(
+                            jump_model,
+                            base_name = "PiecewiseLinearBlockReserveOffer_$(device_type)_{$(service_name), $(dev_name), $(k), $(t)}",
+                            lower_bound = 0.0,
+                        )
+                    JuMP.@constraint(jump_model, v <= breakpoints[k + 1] - breakpoints[k])
```
luke-kiernan, line 78: breakpoints are floats: could create upper bound inside `@variable`

src/services_models/reserves.jl
```diff
@@ -104,41 +104,47 @@ Add variables for ServiceRequirementVariable for StepWiseCostReserve
...
+    t::Int,
     extra::Int,
-)
-    acc = IOM.get_hinted_aff_expr(length(vars) + extra)
-    for v in vars
-        JuMP.add_to_expression!(acc, v)
+) where {
+    U <: Union{Vector{D}, IS.FlattenIteratorWrapper{D}},
+} where {D <: PSY.Component}
+    acc = IOM.get_hinted_aff_expr(length(contributing_devices) + extra)
```
luke-kiernan, line 146: organizational nitpick: `get_hinted_aff_expr` is non-exported, but the only places that call it are in POM. Move to POM or export it

src/services_models/reserves.jl
```diff
@@ -470,47 +483,83 @@ function add_constraints!(
...
+    model::ServiceModel{S, SR},
 ) where {
     S <: Union{PSY.ReserveDemandCurve, PSY.ReserveDemandTimeSeriesCurve},
     SR <: StepwiseCostReserve,
 }
+    # Demand side: price the endogenous ServiceRequirementVariable by the (decremental) ORDC/ASDC
+    # demand curve. `objective_function_multiplier(ServiceRequirementVariable, StepwiseCostReserve)`
+    # is -1, so this enters the objective as a benefit.
     add_reserves_variable_cost!(container, ServiceRequirementVariable, service, SR)
+    # Supply side: price each contributing device's reserve award by its own AS offer curve
```
luke-kiernan: comment clarity/terminology consistency: I had to look up ASDC and AS here

src/services_models/services_constructor.jl
```diff
@@ -379,19 +419,25 @@ function construct_service!(
...
-    service = PSY.get_component(SR, sys, name)
-    !PSY.get_available(service) && return
-    contributing_services = PSY.get_contributing_services(service)
-
-    add_constraints!(
+    groups = collect(get_available_components(model, sys))
+    # Dense group-indexed requirement container, built once over all groups of the type.
+    add_constraints_container!(
         container,
         RequirementConstraint,
```
luke-kiernan, line 426: pre-existing, but super generic name. `ReserveRequirementConstraint` would be clearer

src/services_models/services_constructor.jl
luke-kiernan, line 1: Maybe have AI do a quick pass on this file for low-hanging fruit for code deduplication. Pre-existing but these `constructs_device!` calls all look fairly similar...

luke-kiernan (changes requested):

src/services_models/reserve_offers.jl
rodrigomha: This would require to change most of our market bid pipeline. I think we decided to keep the constraint approach since before, so if we want to refactor this using this we will need to change everything. We could open an issue for this refactor if we want.

src/services_models/reserves.jl
rodrigomha: This sounds right. I will open an issue for later on this since we don't want to touch IOM for now.

src/services_models/services_constructor.jl
rodrigomha: I agree with this, but since ConstraintKeys usually come with both the constraint and device/service type is enough to handle this. I will defer for later a rename of old stuff for now.

src/services_models/reserve_offers.jl
luke-kiernan: If it's something you already considered, then sure, leave it.

I would like to know, though: is the "create upper bound afterward" choice a correctness thing, or just a code uniformity thing?

rodrigomha: I think is highly likely that is code uniformity and how PSI was done before. To create the bounds would require to have access to the curves and probably a specific set-up. Since now we have IOM to create the variables for delta_pwl it is possible to do it, but our approach just follows how it was done before (and how IOM was ported from PSI).

I don't think is correct or not adding it later or in the variable. It is possible that solvers will just transform the constraint internally anyways also. It would be fun to compare solver performance if we do this change.

docs/src/reference/optimization_container_axes.md
rodrigomha: Addressed this in https://github.com/Sienna-Platform/PowerOperationsModels.jl/pull/206/changes/d6ffa2a49c2fe6f2d0c8a1f9166b164d35d025ab

src/common_models/add_expressions.jl
rodrigomha: Agree with this. Changed it in https://github.com/Sienna-Platform/PowerOperationsModels.jl/pull/206/commits/b8cf4cb43523df000f070b4c697006dc6e99a861

src/common_models/add_parameters.jl
rodrigomha: This is true and a bug.  If you do `IS.add_time_series!(sys, [r1, r2], forecast)` gives both reserves the same time series UUID, which is a supported, but will break the build for the service. Working on fixing this.

src/core/problem_template.jl
rodrigomha: Yeah, pretty bad. Trimmed here: https://github.com/Sienna-Platform/PowerOperationsModels.jl/pull/206/commits/454542afe3f79bc7d00a16868078ad3fcdd3cf5b

src/services_models/reserves.jl
rodrigomha: Yeah, I decided to follow the approach in PSY. If an acronym is used in a file it is always defined at the first docstring it appears in that file. The updates are in this commit: https://github.com/Sienna-Platform/PowerOperationsModels.jl/pull/206/commits/c24fe46008753cd298d3b237a0e8ee73b294bfaf

src/services_models/services_constructor.jl
rodrigomha: I opened an issue here: https://github.com/Sienna-Platform/PowerOperationsModels.jl/issues/223

It is clear that we can do that, but we should do it in another PR after the PSY Reserve refactor is done.

src/services_models/reserves.jl
rodrigomha: Open here: https://github.com/Sienna-Platform/PowerOperationsModels.jl/issues/224

src/services_models/reserve_group.jl
rodrigomha: You are right here. I went back and forth with AI on this one. A better solution is to just that store a vector of the variables for every time step, and then simply add the variables all together.  Basically:

> Each iteration was also building a fresh (String, Int) tuple and hashing the string to probe the Dict, and recomputing PSY.get_name(r) inside the time loop. The more interesting part: the (service_name, t) key wasn't buying anything. For a given t the loop visits every contributing service and folds every variable into one sum, so the service dimension collapses immediately, and _group_member_variables has already filtered to group members via member_names. The lookup was doing work whose answer is always "all of them."

With this, I just changed the `_group_member_variables` to just return `Vector{Vector{VariableRef}}`  (of the member variables that are added together at each timestep). A relevant concern is that the order within the bucket is not deterministic and will depend on how `contributing_services` presents them. Since we are adding it all together we don't really care, but if we need it for other thing later we may need to revisit this. I can't think of anything right now of how having the older Dict is relevant for these types of Grouped Services.

See  https://github.com/Sienna-Platform/PowerOperationsModels.jl/pull/206/commits/c88704e622afa13d0520358be5a351434f461384

src/common_models/add_parameters.jl
luke-kiernan: There's enough going on this this PR already. Resolving this comment.

src/common_models/add_parameters.jl
rodrigomha: Addressed this issue to re-use the UUID key rather than name to resemble more the Devices stuff using the dedup. I think using the names was also possible, but probably keeping it closer to devices is a better plan.

Addressed in: https://github.com/Sienna-Platform/PowerOperationsModels.jl/pull/206/commits/b0c5d862005438353e4900f994544559971c6dcc

src/common_models/add_parameters.jl
rodrigomha: Yeah, I'll open an issue for this to keep track.

jd-lara (approved):

PR #213, opened 2026-07-29
psy6 loadability

jd-lara:


src/utils/psy_utils.jl
```diff
@@ -31,3 +31,20 @@ function get_available_turbines(
         PSY.get_downstream_turbines(d),
     )
 end
+
+# psy6: a TwoWindingTransformer's arc and ratings live on its TransformerCircuit, not on
+# the parent device — PSY defines no parent-level `get_arc` / `get_rating` for it. Read
```
m-bossart, line 36: It seems to me like PSY should actually define `get_arc` and `get_rating` on the parent `TwoWindingTransformer`

PR #221, opened 2026-08-05
fix hydro reserve up/down mixing

luke-kiernan:
While I was asking AI questions about an unrelated PR, it noticed a copy-paste bug. Patch it and add a basic regression test.

### DIFF
--- a/src/static_injector_models/hydro_generation.jl
+++ b/src/static_injector_models/hydro_generation.jl
@@ -2293,7 +2293,7 @@ function calculate_aux_variable_value!(
             else
                 served_regup = 0.0
             end
-            if has_container_key(container, HydroServedReserveUpExpression, typeof(d))
+            if has_container_key(container, HydroServedReserveDownExpression, typeof(d))
                 served_regdn = jump_value(
                     get_expression(container, HydroServedReserveDownExpression, T)[
                         name,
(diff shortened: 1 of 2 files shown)


src/static_injector_models/hydro_generation.jl
```diff
@@ -2293,7 +2293,7 @@ function calculate_aux_variable_value!(
             else
                 served_regup = 0.0
             end
-            if has_container_key(container, HydroServedReserveUpExpression, typeof(d))
+            if has_container_key(container, HydroServedReserveDownExpression, typeof(d))
```
acostarelli, line 2296: How often are we doing this sort of guarding in aux variable  calculations? I wonder if we could make a helper here.

test/test_device_hydro_constructors.jl
```diff
@@ -513,6 +513,89 @@ end
...
+        IOM.get_expressions(container),
+        IOM.ExpressionKey(HydroServedReserveUpExpression, HydroDispatch),
+    )
+    POM.calculate_aux_variable_value!(
+        container,
+        IOM.AuxVarKey{HydroEnergyOutput, HydroDispatch}(""),
+        c_sys5_hy,
+    )
+
+    @test all(isapprox.([aux[hy_name, t] for t in time_steps], expected; atol = 1e-8))
```
acostarelli, line 594: I'm confused how this test passes, shouldn't served_regup in hydro_generation now be 0? Or was it 0 before in this test, and so this is redundant?

It also seems strange to be testing what happens when you manually delete a container key, which isn't something we'd ever do, but maybe I'm missing the narrative of this test.

jd-lara (approved):

PR #233, opened 2026-08-12
network matrix consolidation

jd-lara:
This PR consolidated the use of a single network information entering the models. The purpose is to avoid having more than one factorization, using a more coherent interface for the reductions and simplifiying the exceptions code. 

It will likley need refinements later 

requires https://github.com/Sienna-Platform/InfrastructureOptimizationModels.jl/pull/152

### DIFF
--- a/src/network_models/instantiate_network_model.jl
+++ b/src/network_models/instantiate_network_model.jl
@@ -273,9 +273,8 @@ end
 
 # A prebuilt source fixed the reduction before the build, so its Ybus is reproduced from
 # the source's own reduction rather than from the model. The result is checked against
-# that reduction: anything the spec cannot reproduce (a non-default zero-impedance
-# threshold, say) must fail loudly rather than hand downstream code a Ybus describing a
-# different network than the matrices do.
+# that reduction: anything the spec cannot reproduce must fail loudly rather than hand
+# downstream code a Ybus describing a different network than the matrices do.
 function _prebuilt_ybus(
     source::IOM.AbstractNetworkSource,
     reduction::PNM.NetworkReductionData,
--- a/src/network_models/instantiate_network_model.jl
+++ b/src/network_models/instantiate_network_model.jl
@@ -218,6 +218,10 @@ function _reduced_ybus!(
     return ybus
 end
 
+# Every PTDF-family matrix wraps one of these, so the tolerance and uuid are set once.
+_factor_core(ybus::PNM.Ybus, sys::PSY.System) =
+    PNM.VirtualFactorCore(ybus; tol = PTDF_ZERO_TOL, system_uuid = IS.get_uuid(sys))
+
 #=
 The one source-aware Ybus resolution, dispatched on the source so every formulation
 family gets it. A source that derives the network from the system applies the build's
@@ -235,22 +239,17 @@ function _source_ybus(
     return _build_ybus(source, sys, exceptions)
 end
 
-function _source_ybus(
-    source::PrebuiltMatrixSource,
-    sys::PSY.System,
-    exceptions::Vector{Int},
-)
-    reduction = PNM.get_network_reduction_data(PNM.get_core(get_matrix(source)))
-    _validate_prebuilt_exceptions(reduction, exceptions)
-    return _prebuilt_ybus(source, reduction, sys)
-end
+_prebuilt_reduction(source::PrebuiltMatrixSource) =
+    PNM.get_network_reduction_data(PNM.get_core(get_matrix(source)))
+_prebuilt_reduction(source::PrebuiltCoreSource) =
+    PNM.get_network_reduction_data(get_core(source))
 
 function _source_ybus(
-    source::PrebuiltCoreSource,
+    source::Union{PrebuiltMatrixSource, PrebuiltCoreSource},
     sys::PSY.System,
     exceptions::Vector{Int},
 )
-    reduction = PNM.get_network_reduction_data(get_core(source))
+    reduction = _prebuilt_reduction(source)
     _validate_prebuilt_exceptions(reduction, exceptions)
     return _prebuilt_ybus(source, reduction, sys)
 end
@@ -380,7 +379,7 @@ function IOM.instantiate_network_model!(
     ybus = _reduced_ybus!(model, sys, exceptions)
     IOM.set_network_data!(
         model,
-        YbusNetworkData(ybus, deepcopy(PNM.get_network_reduction_data(ybus))),
+        YbusNetworkData(ybus, _owned_reduction(ybus)),
     )
     _finalize_network_reduction!(model, branch_models, number_of_steps)
     return
@@ -399,13 +398,9 @@ function IOM.instantiate_network_model!(
     _validate_network_and_branches(model, branch_models, sys)
     exceptions = _collect_reduction_exceptions(sys, model, branch_models)
     ybus = _reduced_ybus!(model, sys, exceptions)
-    reduction = deepcopy(PNM.get_network_reduction_data(ybus))
+    reduction = _owned_reduction(ybus)
     if IOM._template_has_outage_aware_branch(branch_models)
-        core = PNM.VirtualFactorCore(
-            ybus;
-            tol = PTDF_ZERO_TOL,
-            system_uuid = IS.get_uuid(sys),
-        )
+        core = _factor_core(ybus, sys)
         IOM.set_network_data!(
             model,
             DCPNetworkData(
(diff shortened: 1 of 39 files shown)


jd-lara (commented):
One thing to note is that I realized because Claude fucked up that there is a bug in the modeling of constant impedance loads, I'll leave that for a later PR

src/network_models/network_data.jl
```diff
@@ -0,0 +1,83 @@
...
+matrix it never derived.
+
+The reduction data is a deepcopy: POM extends it with `populate_branch_maps_by_type!`
+and must never write into a core that the caller may have supplied or shared. The
+copy may only ever be EXTENDED — its bus and arc mapping must stay identical to the
+core's, because the MODF is indexed with arcs resolved through this copy.
+=#
+
+"""Ybus-only families: no factorization, no sensitivity matrices."""
+struct YbusNetworkData <: IOM.AbstractNetworkData
```
m-bossart, line 23: Why do we need to store both the matrix and the NetworkReductionData in these structs when the matrix always has the NetworkReductionData internally?

src/network_models/network_sources.jl
```diff
@@ -0,0 +1,110 @@
...
+"""
+Reuse a `VirtualPTDF` the caller already built, including its populated row cache.
+Its own reduction becomes the build's reduction; the reduction exceptions derived
+from the template are not applied, because the matrix already exists.
+
+Only a `VirtualPTDF` is accepted: it carries the factorization core, so a MODF
+sharing its reduction can be derived from it. A dense `PNM.PTDF` carries no core and
+is deliberately not a valid source.
+"""
+struct PrebuiltMatrixSource{M <: PNM.VirtualPTDF} <: IOM.AbstractNetworkSource
```
luke-kiernan, line 31: Why the type parameter here? `VirtualPTDF` doesn't have subtypes.

src/network_models/network_data.jl
```diff
@@ -0,0 +1,83 @@
...
+
+"""PTDF families: a PTDF, plus an optional MODF, both wrapping one factorization core."""
+struct PTDFNetworkData{P, M} <: IOM.AbstractNetworkData
+    matrix::P
+    contingency_matrix::M
+    reduction::PNM.NetworkReductionData
+end
+
+# The no-contingency constructors omit the field rather than storing `nothing`, so
+# `has_contingency_matrix` is a dispatched predicate on a concrete type.
```
luke-kiernan: Huh what....does this actually good sense? Can't tell.

luke-kiernan: From conversations with Claude:
1. "omit the field" is more like "omit the argument:" constructor ergonomics, not struct layout.
2. "so...is a dispatched predicate on a concrete type:" `Nothing` is a concrete type. Dispatching on `::DCPNetworkData{Nothing}` would work just as well.

src/network_models/network_sources.jl
```diff
@@ -0,0 +1,110 @@
...
+    matrix::M
+end
+
+"""
+Reuse a factorization core the caller already built — the cheapest reuse, since the
+PTDF and MODF wrappers are derived from it without re-factorizing the ABA matrix.
+"""
+struct PrebuiltCoreSource <: IOM.AbstractNetworkSource
+    core::PNM.VirtualFactorCore
```
luke-kiernan: Why not just pass around `core::PNM.VirtualFactorCore` directly? Why do we need the outer `PrebuiltCoreSource` wrapper?

src/network_models/reduction_exceptions.jl
```diff
@@ -0,0 +1,177 @@
...
+# AreaInterchange <: PSY.Branch but connects two Areas, not two buses; it has no
+# arc, so the PSY.Branch method above would error on PSY.get_arc. Not reachable
+# today (modeled_branch_types entries are filtered to <: PSY.ACTransmission
+# before reaching _push_component_buses!, and AreaInterchange isn't one), but
+# guard it explicitly so a future caller can't hit that MethodError.
+function _push_component_buses!(::Set{Int}, ::PSY.AreaInterchange)
+    return
+end
+
+# Fallback for monitored/outaged component types with no bus-pinning rule. Reached
```
luke-kiernan: Loooong comment.

luke-kiernan (commented):
First pass. Again, didn't look at the tests or Claude files.

Why are some of these pieces here, not elsewhere? For example, we could move the definitions of both `AbstractNetworkSource` and `NetworkReductionSpec` into PNM. Then the rest of the `network_sources.jl` file could land there, too. So why don't we do that?

src/network_models/network_data.jl
```diff
@@ -0,0 +1,83 @@
...
+"""DCP: a Ybus, plus a MODF derived from that same Ybus when the template is security constrained."""
+struct DCPNetworkData{M} <: IOM.AbstractNetworkData
+    ybus::PNM.Ybus
+    contingency_matrix::M
+    reduction::PNM.NetworkReductionData
+end
+
+"""PTDF families: a PTDF, plus an optional MODF, both wrapping one factorization core."""
+struct PTDFNetworkData{P, M} <: IOM.AbstractNetworkData
+    matrix::P
```
m-bossart, line 37: Shouldn't P be restricted here, and in the constructor below to `VirtualPTDF` because the `PTDF` does not have the factorization core

src/network_models/instantiate_network_model.jl
```diff
@@ -347,99 +187,148 @@ function _validate_network_and_branches(
...
+        end
     end
-    model.network_reduction = deepcopy(PNM.get_network_reduction_data(ybus))
-    #if !isempty(model.network_reductionget_net_reduction_data)
-    # TODO: Network reimplement this when it becomes necessary. We don't have any
-    # reductions that are incompatible right now.
-    # check_network_reduction_compatibility(T)
-    #end
-    return
+    _consolidate_device_model_outages_with_modf!(branch_models, modf)
```
m-bossart, line 345: Do we still need this? It seems like it is unreachable now and can be deleted.

src/network_models/network_sources.jl
m-bossart: I think so it can be defined here in POM.

src/network_models/instantiate_network_model.jl
```diff
@@ -461,50 +350,58 @@ function _finalize_network_reduction!(
 end
 
 #################################################################################
-# AreaBalanceNetworkModel
+# Ybus-only families (ACP, ACR, IVR, LPACC, NFA, CopperPlate, AreaBalance, ...)
```
m-bossart, line 368: It seems like CopperPlate and AreaBalance should  never apply a non-default network reduction. A reduction could in theory move an injector from one area to another based on the remapping. This might be a reason to keep a separate method for those two from the other Ybus only cases.

m-bossart: We do pin buses as irreducible if they have cross area boundaries in PNM, but only if there are AreaInterchanges in the system. Either way, I think we should have a separate method here and error if the user passes network reductions and tries to use copperplate. It doesn't buy anything and tells the user their inputs aren't being used.

src/network_models/network_data.jl
```diff
@@ -5,9 +5,15 @@ dispatch instead of an isnothing check, and a family cannot accidentally read a
...
-core's, because the MODF is indexed with arcs resolved through this copy.
+and must never write into a core that the caller may have supplied or shared. That
+function only ever accumulates — it never clears — so sharing it would leak one
+template's branch maps into the next build off the same core. The copy may only ever be
+EXTENDED: its bus and arc mapping must stay identical to the core's, because the MODF is
+indexed with arcs resolved through this copy.
+
+Consequence: `PNM.get_network_reduction_data(matrix)` and `get_reduction(network_data)`
+are the same reduction but not the same object, and only the latter carries the branch
+maps. Read the reduction off the container, never off the matrix.
```
luke-kiernan, line 16: > Consequence: `PNM.get_network_reduction_data(matrix)` and `get_reduction(network_data)`
are the same reduction but not the same object, and only the latter carries the branch
maps. Read the reduction off the container (2nd option), not the matrix (1st option).

Yikes that seems like a recipe for bugs and confusion. We have two objects of the same type, representing the same thing, that are different...imo that's a sign that there's something amiss with the design or usage of the object. Sounds like the network data's reduction only adds more stuff: go to 2 different structs, one containing the other?

m-bossart: I think this is a good point. The branch maps that are organized by type are currently stored within the NetworkReductionData object, but only populated and consumed from POM. We should keep the logic for building them in PNM, but move the maps to a POM owned struct. I'm working on PRs in POM (on top of this one) and PNM to make this change. There is also a trivial change required in PFs for one of the tests.

m-bossart (approved):
@luke-kiernan Your comment is a good one and I agree. I opened up an issue to track the improvements that are needed to keep a single source of truth and prevent the divergence you mentioned. I think we can merge this now and follow up with additional improvements.

PR #235, opened 2026-08-12
Migrate services to the psy6 reserve tree (OnlineReserve/OfflineReserve/GroupReserve)

rodrigomha:
Reserve dispatch re-expressed on the psy6 reserve tree, with distinctions recovered from state instead of struct types: direction and non-spin via `reserve_traits.jl`, ORDCs via the demand curve on `variable`, static vs time-varying requirements via the attached series. Includes the degenerate-demand supply-only skip (services can serve a `GroupReserve`), the `GroupRangeReserve` rename, instance-derived container metas, and two independent fixes in their own commits: the `Symbol(T)`-vs-`nameof` silent service-device wiring skip under parallel test workers, and the units-mandatory `get_max_shunt_current` read.

Full suite green locally against the local psy6 stack: 105891/105891.

**Scope note:** this PR is the refactor only. The elastic group ORDC (`GroupStepwiseCostReserve`) and load reserve provision (`PowerLoadDispatch` up/down reserves with the costless-load guard) are NEW features, not migrations of existing behavior, and will land in a follow-up PR together with the market integration test and their `formulation_library.md` entries.

Depends on the PSY reserve changes (`jd/schema_matching`), Sienna-Platform/PowerSystemsTestData#124, and Sienna-Platform/PowerSystemCaseBuilder.jl#212 - CI cannot pass until those merge and `[sources]` resolve.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,7 @@ docs/site/
 Manifest.toml
 .vscode
 *.h5
+.claude/plans
 
 ################################################################################
 #                              Operating systems                               #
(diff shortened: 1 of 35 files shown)


jd-lara (approved):
This PR is mostly renaming the reserves to consider the new psy6 constructions. I'll merge to restore the tests

PR #236, opened 2026-08-13
Group-stepwise ORDC, load reserve provision, and OfflineReserve ORDC supply

rodrigomha:
Stacked on #235 (base branch `rh/reserve_refactor`); the diff shows only this feature's commits.

New features (not refactor items) re-expressed on the psy6 reserve tree:

- **`GroupStepwiseCostReserve`**: elastic group ORDC. One dense `ServiceRequirementVariable` per group, a clearing constraint `sum(member awards) >= requirement variable`, and the group's demand curve (static or time series) priced through the existing delta-PWL path. Mis-paired `ServiceModel`s (`GroupReserve` with a non-group formulation and vice versa) fail at declaration.
- **Load reserve provision under `PowerLoadDispatch`**: the inverse of a generator. Up reserve is committed shed (`P - r_up >= 0`), down reserve is committed extra consumption (`P + r_down <= forecast`), gated on an attached service model. A costless load selling reserves errors since nothing pins its consumption.
- **`OfflineReserve` (non-spin) as ORDC supply from storage and loads**: non-spinning is upward-only, so it routes like an up reserve everywhere a device supplies it (new `UP_RESERVE` union). Previously only thermal was wired: the storage reserve-balance multipliers MethodError'd, two storage coverage branches silently skipped the service, complete coverage hit an `@assert false`, and `_modify_device_model!` no-opped every `OfflineReserve` model. The no-op is now scoped to `NonSpinningReserve`, whose awards ride `ReservePowerConstraint` instead of the device range expressions.
- **Integration test**: energy + reserve co-clearing with an elastic service, a stepwise group, and per-resource offers from thermal, storage, and load participants.

Tests live in existing files: group-stepwise testsets in `test_services_constructor.jl`; market integration, load provision, and the `OfflineReserve` ORDC scenarios in `test_device_reserve_offers.jl`; the storage non-spin coverage testset in `test_storage_device_models.jl`.

Full suite green: 106708/106708. Docs build green.

### DIFF
--- a/src/common_models/add_to_expression.jl
+++ b/src/common_models/add_to_expression.jl
@@ -1957,11 +1957,7 @@ function add_to_expression!(
     model::DeviceModel{V, W},
     network_model::NetworkModel{X},
 ) where {
-    T <: Union{
-        ActivePowerRangeExpressionUB,
-        ActivePowerRangeExpressionLB,
-        ActivePowerRangeExpressionOnlineUB,
-    },
+    T <: Union{ActivePowerRangeExpressionUB, ActivePowerRangeExpressionLB},
     U <: VariableType,
     V <: PSY.Device,
     W <: AbstractDeviceFormulation,
@@ -2001,27 +1997,10 @@ function add_to_expression!(
     end
     expression = get_expression(container, T, V)
     time_steps = get_time_steps(container)
-    # Online up-reserves also occupy the online-only band row when a device carries an
-    # OfflineReserve (the OnlineUB expression exists only in that case); offline awards
-    # live solely in the shared UB band, freed from the commitment gate by
-    # OfflineReserveBandConstraint.
-    online_ub =
-        if !(service isa PSY.OfflineReserve) &&
-           has_container_key(container, ActivePowerRangeExpressionOnlineUB, V)
-            get_expression(container, ActivePowerRangeExpressionOnlineUB, V)
-        else
-            nothing
-        end
     for d in devices, t in time_steps
         name = PSY.get_name(d)
         add_proportional_to_jump_expression!(
             expression[name, t],
-            variable[(service_name, name, t)],
-            1.0,
-        )
-        online_ub === nothing && continue
-        add_proportional_to_jump_expression!(
-            online_ub[name, t],
             variable[(service_name, name, t)],
             1.0,
         )
@@ -2507,11 +2486,7 @@ function add_to_expression!(
     devices::IS.FlattenIteratorWrapper{V},
     model::DeviceModel{V, W},
 ) where {
-    T <: Union{
-        ActivePowerRangeExpressionUB,
-        ActivePowerRangeExpressionLB,
-        ActivePowerRangeExpressionOnlineUB,
-    },
+    T <: Union{ActivePowerRangeExpressionUB, ActivePowerRangeExpressionLB},
     U <: OnStatusParameter,
     V <: PSY.Device,
     W <: AbstractDeviceFormulation,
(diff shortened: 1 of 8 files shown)


jd-lara:
@rodrigomha this PR needs to be rebased

rodrigomha:
@jd-lara rebased onto main (post #235 squash-merge): the 7 feature commits replayed clean with zero conflicts (`git rebase --onto main <235-head>`), full suite green on the rebased branch (106708/106708). Base is already `main`.

rodrigomha:
Done with the rebase

rodrigomha:
CI status after the rebase, fully diagnosed:

- **Fixed on this branch** (`518ae0e`, `f0fda18`): environment resolution. PSY psy6 now depends on the unregistered OpenAPI-models packages, and PSB psy6 on PowerTableDataParser/PowerFlowFileParser - Pkg ignores `[sources]` of non-root projects, so POM's root/test/docs environments now pin them directly. Docs, build, and comparison jobs are green.
- **Remaining test-job failures are upstream PSY psy6 bugs**, reproduced and fixed locally (pending review before push): (1) the generated `HydroPumpTurbine` exporter calls `_turbinepump_po`, which was never added to the hand-written export helpers; (2) `DOCUMENT_PLAN` imports `HydroPumpTurbine` after `HydroReservoir`, which resolves its unit ids -> unresolved-id error; (3) `FuelCurve.fuel_cost::TimeSeriesKey` has no OpenAPI serde, which breaks PSB's bundle cache for `c_sys5_re_fuel_cost` (fixed via the schema's string variant + a document-level key-resolution pass).
- With those PSY fixes applied locally: PSY `test_openapi_document` 448/448, POM full suite **106708/106708**. No further POM commits needed - the pins float on psy6.
- Separate flag: PSB's new bundle cache write is non-atomic; parallel first-builds can read a half-written `system.json` (`UnexpectedEOF`) - the old first-serialization race in a new spot. Worth a write-to-temp+rename in PSB.

rodrigomha:
Restacked on #233 per the network-consolidation sequencing: base is now `jd/network_matrix_consolidation` and the branch is rebased onto it (`4e28ede`), so the diff shows only this PR's commits.

- The 7 feature commits replayed clean. Of the two CI environment-pin commits, the PSB transitive-deps one was dropped entirely and the OpenAPI one shrank to docs-env pins only - #233 already carries the same root/test pins (same non-root `[sources]` traps, fixed independently there).
- Note #233 pins IOM at `jd/network-sources`; this stack inherits that.
- Validation on the stacked branch with latest psy6 (which now includes the `_turbinepump_po` and hydro DOCUMENT_PLAN fixes): compile + zero method ambiguities, and the full suite **106951/106951**. The one remaining upstream gap for CI is the `FuelCurve.fuel_cost` time-series serde in PSY (fix staged locally, pending review) - without it the thermal fuel-cost test fails via PSB's bundle cache.

src/common_models/add_to_expression.jl
```diff
@@ -2413,6 +2411,74 @@ function add_to_expression!(
     return
 end
 
+# Load up-reserve is committed shed: LB = P - Σ r_up, constrained >= 0. Generators route
+# ReserveUp to the UB expression, so `V <: PSY.ElectricLoad` cannot shadow them.
```
luke-kiernan, line 2415: I don't follow the 2nd sentence of this comment

src/core/reserve_traits.jl
```diff
@@ -31,11 +31,18 @@ struct ChargeSide <: ReserveSide end
 
 """
 Direction of a reserve. `OfflineReserve` (non-spinning) has no direction type parameter and is
-upward-only in every US market, so it maps to [`PSY.ReserveUp`](@ref).
+upward-only in every US market, so it maps to `PSY.ReserveUp`.
```
luke-kiernan, line 34: Lost `(@ref)`. purposeful or AI?

luke-kiernan (commented):
Looks fine, no issues at a glance. However, you should probably have others review it too--I'm not very familiar the distinctions between these differennt types of reserves...

src/common_models/add_to_expression.jl
rodrigomha: This is mostly that the bound for ReserveUp for Generators is done using the Upper Bound since ReserveUp is provided by increasing generation, while for loads ReserveUp is provided by curtailing load, and hence bounded via LowerBound.

src/core/reserve_traits.jl
rodrigomha: It's probably AI that did it because of an issue with the docs, we need to figure out the docs by the end of it.

jd-lara (approved):

PR #239, opened 2026-08-15
Add tap controls; refactor AC_branches.jl

acostarelli:
Add support for tap control objectives, and large refactoring of AC_branches.jl


test/includes.jl
```diff
@@ -7,7 +7,7 @@ using InfrastructureSystems
 import InfrastructureSystems: TableFormat
 using PowerNetworkMatrices
 import PowerSystemCaseBuilder: PSITestSystems
-using PowerFlows
+#using PowerFlows
 using DataFramesMeta
```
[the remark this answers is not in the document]
acostarelli: This was running into precompilation issues. Maybe something upstream changed.

src/network_models/instantiate_network_model.jl
```diff
@@ -162,78 +165,77 @@ function _add_outage_monitored_irreducible_buses!(
...
+function _controllable_transformers_irreducbile_buses(branch_models::BranchModelContainer)
+    irreducible_buses = Set{Int64}()
+    for model in values(branch_models)
+        _control_enabled(model) || continue
+        for transformer in get_device_cache(model)
+            for circuit in PSY.get_circuits(transformer)
+                _control_enabled(circuit) || continue
+                _push_component_buses!(irreducible_buses, circuit)
+                push!(irreducible_buses, PSY.get_regulated_bus_number(circuit))
+            end
```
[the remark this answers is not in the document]
acostarelli: I think the convention we're going to go for now is the regulated bus number must always be a valid bus number, so 0 doesn't carry special importance.

That said, I agree maybe we shouldn't be pinning the regulated bus on all controls.

src/ac_transmission_models/AC_branches.jl
```diff
@@ -1113,52 +1204,25 @@ function _branch_rating_entries(
...
-# reduction folds a FIXED device setting into the merged π-parameters. A controlled
-# branch absorbed by a network reduction is a modeling conflict the user must resolve,
-# not something to silently approximate.
 function _validate_controlled_branch_not_reduced(
     network_model::NetworkModel,
-    devices::IS.FlattenIteratorWrapper{T},
-    formulation_name::String,
+    ::Type{T},
+    controlled_names,
 ) where {T <: PSY.ACTransmission}
```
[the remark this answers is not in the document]
acostarelli: transformer_models.jl isn't used. I will remove it when I implement phase shifting.

I believe a parallel merge is the only merge that can happen; series merges are blocked by specifying irreducible buses.

jd-lara:
@acostarelli I stacked this on top of the PR with the network changes

src/ac_transmission_models/AC_branches.jl
```diff
@@ -271,12 +312,62 @@ function add_variables!(
...
+    container::OptimizationContainer,
+    model::DeviceModel{U, F},
+    devices::IS.FlattenIteratorWrapper{U},
+    network_model::NetworkModel,
+) where {
+    U <: _TRANSFORMERS,
+    F <: AbstractBranchFormulation,
+}
+    get_attribute(model, ENABLE_CONTROLS_KEY) === true || return
+    _warn_tap_control_nonconvexity(network_model)
```
acostarelli: This gets called before we even know if any tap controls exist.

src/ac_transmission_models/AC_branches.jl
```diff
@@ -271,12 +312,62 @@ function add_variables!(
...
+function _add_tap_control_variables!(
+    container::OptimizationContainer,
+    model::DeviceModel{U, F},
+    devices::IS.FlattenIteratorWrapper{U},
+    network_model::NetworkModel,
+) where {
+    U <: _TRANSFORMERS,
+    F <: AbstractBranchFormulation,
+}
+    get_attribute(model, ENABLE_CONTROLS_KEY) === true || return
```
acostarelli: Use the helper.

src/ac_transmission_models/AC_branches.jl
```diff
@@ -1172,7 +1236,46 @@ _dc_phase_shift(branch::PNM.AbstractReductionAggregate, nr::PNM.NetworkReduction
 _dc_phase_shift(branch::PSY.ACTransmission, ::PNM.NetworkReductionData) =
     PNM.get_series_phase_shift(branch)
 
-function _branch_geometry(
+_get_circuit(b::_TRANSFORMERS) = PSY.get_circuit(b)
```
acostarelli: I don't think this works for 3W. Also, we need to add more coverage for 3W.

src/ac_transmission_models/AC_branches.jl
```diff
@@ -1694,35 +1766,185 @@ function add_constraints!(
...
-            vsq_fr = vr_fr^2 + vi_fr^2
-            vsq_to = vr_to^2 + vi_to^2
-            vv_cos = vr_fr * vr_to + vi_fr * vi_to
-            vv_sin = vi_fr * vr_to - vr_fr * vi_to
-            _add_tap_acr_flow!(
-                jump_model, cons_pft, cons_qft, cons_ptf, cons_qtf, pft, qft, ptf, qtf,
-                name, t, vsq_fr, vsq_to, vv_cos, vv_sin, coef, tm,
-                _slack_term(slacks.p_ft, name, t),
-                _slack_term(slacks.q_ft, name, t),
+            vp = _voltage_products(container, network_model, T, name, from_bus, to_bus, t)
```
acostarelli: This doesn't need to happen in the time loop. This method can return `DAAs{OneTo}` instead.

src/ac_transmission_models/AC_branches.jl
```diff
@@ -1694,35 +1766,185 @@ function add_constraints!(
...
+_voltage_magnitude(container, name, ::NetworkModel{ACPNetworkModel}) =
+    get_variable(container, VoltageMagnitude, PSY.ACBus)[name, :]
+_voltage_magnitude(
+    container,
+    name,
+    ::NetworkModel{<:Union{ACRNetworkModel, IVRNetworkModel}},
+) =
+    JuMP.@expression(
+        get_jump_model(container),
+        [t in get_time_steps(container)],
```
acostarelli: It's probably beneficial to ensure this specializes as `OneTo` but I'm not sure it does.

src/ac_transmission_models/AC_branches.jl
```diff
@@ -1694,35 +1766,185 @@ function add_constraints!(
...
+            _voltage_controlled(device_model, circuit) || continue
+
+            bus = PSY.get_bus(sys, PSY.get_regulated_bus_number(circuit))
+            bus_name = PSY.get_name(bus)
+            bus_limits = PSY.get_voltage_limits(bus)
+            ctl_limits = PSY.get_controlled_quantity_limits(circuit)
+            # TODO: temporary pending PSY#1755
+            circuit_name = _circuit_arc_name(d, circuit, i)
+            (bus_limits.min <= ctl_limits.min <= ctl_limits.max <= bus_limits.max) || error(
+                "Bus limits for $bus_name disagree with control limits for circuit $circuit_name.",
```
acostarelli: Specify bus voltage limits

src/ac_transmission_models/AC_branches.jl
```diff
@@ -1694,35 +1766,185 @@ function add_constraints!(
...
+    device_model::DeviceModel{T},
+    ::NetworkModel{<:NativeACNetworkModel},
+) where {T <: _TRANSFORMERS}
+    _control_enabled(device_model) || return
+
+    cons = add_constraints_container!(
+        container,
+        ReactivePowerFlowControlConstraint,
+        T,
+        String[],
```
acostarelli, line 1497: Side-thought: we shouldn't have to pass empty vectors to make sparse arrays, just the type alone should do.

src/ac_transmission_models/AC_branches.jl
```diff
@@ -1694,35 +1766,185 @@ function add_constraints!(
...
+        Int[],
+        Int[];
+        sparse = true,
+    )
+    qft = get_variable(container, FlowReactivePowerFromToVariable, T)
+    qtf = get_variable(container, FlowReactivePowerToFromVariable, T)
+
+    time_steps = get_time_steps(container)
+    jump_model = get_jump_model(container)
+    for d in devices
```
acostarelli: This is the second two-layer loop for circuits, maybe refactor out? Since there will likely be more.

acostarelli: As in `_iter_circuits(devices) = iterator of the circuits`

src/ac_transmission_models/AC_branches.jl
```diff
@@ -1869,106 +2091,23 @@ function _entry_angle_limits(geometry, device_by_name::Dict{String, <:PSY.ACTran
...
-            cs_b = cs[name, t]
+################################## IVRNetworkModel branch constraints ##################
 
-            # Shared affine terms reused across the four flow constraints:
-            #   cs_sum = cs + phi_fr + phi_to,  dev_* = 1 + 2·phi_*
-            cs_sum = cs_b + phi_fr + phi_to
-            dev_fr = 1.0 + 2.0 * phi_fr
-            dev_to = 1.0 + 2.0 * phi_to
+_branch_arc(d::PSY.ACTransmission) = PSY.get_arc(d)
+_branch_arc(d::PSY.TwoWindingTransformer) = PSY.get_arc(PSY.get_circuit(d))
```
acostarelli: What about 3W?

src/ac_transmission_models/AC_branches.jl
```diff
@@ -1694,35 +1766,185 @@ function add_constraints!(
...
-            vsq_to = vr_to^2 + vi_to^2
-            vv_cos = vr_fr * vr_to + vi_fr * vi_to
-            vv_sin = vi_fr * vr_to - vr_fr * vi_to
-            _add_tap_acr_flow!(
-                jump_model, cons_pft, cons_qft, cons_ptf, cons_qtf, pft, qft, ptf, qtf,
-                name, t, vsq_fr, vsq_to, vv_cos, vv_sin, coef, tm,
-                _slack_term(slacks.p_ft, name, t),
-                _slack_term(slacks.q_ft, name, t),
+            vp = _voltage_products(container, network_model, T, name, from_bus, to_bus, t)
+            tap = if _tap_controlled(device_model, g_geom)
```
acostarelli: This also doesn't need to happen in the time loop.

src/ac_transmission_models/AC_branches.jl
```diff
@@ -2496,25 +2617,40 @@ function add_constraints!(
...
+    jump_model = get_jump_model(container)
+    tap_var =
+        if has_container_key(container, TapRatioVariable, T)
+            get_variable(container, TapRatioVariable, T)
+        else
+            nothing
+        end
+
+    for g in geoms, t in time_steps
+        angle = va[g.from_name, t] - va[g.to_name, t] - g.shift_dc
```
acostarelli: Build this in @ expression of add_to_expression!

src/ac_transmission_models/AC_branches.jl
```diff
@@ -2496,25 +2617,40 @@ function add_constraints!(
...
+        flow =
```
acostarelli: Is flow a good name here? Claude chose it. Maybe power instead?

src/ac_transmission_models/AC_branches.jl
```diff
@@ -2979,14 +3049,24 @@ function add_constraints!(
...
+            nothing
         end
+
+    for g in geoms, t in time_steps
+        angle = va[g.from_name, t] - va[g.to_name, t] - g.shift_dc
+        cons[g.name, t] =
+            if _tap_controlled(device_model, g)
+                JuMP.@constraint(
+                    jump_model,
+                    pft[g.name, t] * tap_var[g.name, t] == g.b_dc * g.adm.tap * angle
```
acostarelli: Nitpick: reorder

src/ac_transmission_models/AC_branches.jl
```diff
@@ -2496,25 +2617,40 @@ function add_constraints!(
...
+                    p[g.name, t] - slack_ub[g.name, t] + slack_lb[g.name, t]
+                )
+            else
+                p[g.name, t]
+            end
+        cons[g.name, t] =
+            if _tap_controlled(device_model, g)
+                JuMP.@constraint(
+                    jump_model,
+                    flow * tap_var[g.name, t] == g.b_dc * g.adm.tap * angle
```
acostarelli: Nitpick: reorder

src/ac_transmission_models/AC_branches.jl
```diff
@@ -2979,14 +3049,24 @@ function add_constraints!(
...
-            )
+    tap_var =
+        if has_container_key(container, TapRatioVariable, T)
+            get_variable(container, TapRatioVariable, T)
+        else
+            nothing
         end
+
+    for g in geoms, t in time_steps
+        angle = va[g.from_name, t] - va[g.to_name, t] - g.shift_dc
```
acostarelli: Again, build an expression

src/core/constraints.jl
```diff
@@ -200,6 +200,13 @@ struct ReferenceBusConstraint <: ConstraintType end
 """Rectangular-coordinate voltage magnitude bounds: vmin² ≤ vr² + vi² ≤ vmax²."""
 struct VoltageMagnitudeConstraint <: ConstraintType end
 """
+Terminal reactive-flow band for a transformer circuit whose control objective is
```
acostarelli: Claude keeps using this word "band" that I'm not a fan of

src/network_models/instantiate_network_model.jl
```diff
@@ -66,6 +74,170 @@ _assign_subnetworks_to_buses(
...
+           should be protected, add a _push_component_buses! method for it." maxlog = 5
+    return
+end
+
+function _get_irreducible_buses(
+    sys::PSY.System,
+    network_model::NetworkModel,
+    branch_models::BranchModelContainer,
+)
+    @debug "Identifying buses that are irreducible due to monitored components"
```
acostarelli: This debug wasn't here before, I'm not sure it's needed.

src/network_models/instantiate_network_model.jl
```diff
@@ -66,6 +74,170 @@ _assign_subnetworks_to_buses(
...
+            _push_component_buses!(irreducible_buses, component)
+        end
+    end
+    return irreducible_buses
+end
+
+_is_ac_transmission(::DeviceModel{<:PSY.ACTransmission}) = true
+_is_ac_transmission(_) = false
+
+_is_3w(::DeviceModel{PSY.ThreeWindingTransformer}) = true
```
acostarelli: Maybe change to `_is_three_winding`

acostarelli (commented):
Implementation review

test/test_native_dcp_acp_models.jl
```diff
@@ -483,29 +479,20 @@ end
...
-
-    # sys14 = PSB.build_system(PSITestSystems, "c_sys14")
-    # template = get_thermal_dispatch_template_network(NetworkModel(ACPNetworkModel))
-    # set_device_model!(
-    #     template,
-    #     DeviceModel(PSY.TwoWindingTransformer, VoltageControlTap; use_slacks = true),
-    # )
-    # model = DecisionModel(template, sys14; optimizer = ipopt_optimizer)
-    # @test_throws IS.ConflictingInputsError POM.validate_template(model)
+    # slack_spec defaults to NoBranchSlacks, so every pair whose constructors build no
```
acostarelli, line 486: This is not faithful to the original: c_sys14 was dropped.

test/test_power_flow_in_the_loop.jl
```diff
@@ -184,72 +187,12 @@ end
...
 #   - Breaker-switch (DiscreteControlledACBranch)
 #   - HVDCs with DC PowerFlow
 #   - Line active power loss aux variable
 # -----------------------------------------------------------------------------
 
-@testset "AC Power Flow in the loop for PhaseShiftingTransformer" begin
-    #    system = buid_system(PSITestSystems, "c_sys5_uc")
-    #
-    #    line = get_component(Line, system, "1")
-    #    arc = get_arc(line)
```
acostarelli, line 198: I'd kinda prefer keeping PST tests where they are for now. I don't know if these were moved or just deleted.

test/test_native_transformer_tap.jl
```diff
@@ -1,114 +0,0 @@
-#########################################################################################
-# Off-nominal transformer tap under the native network models.
```
acostarelli, line 2: I would also prefer not deleting tests until we know that coverage was properly ported.

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
+#########################################################################################
+# Transformer tap controls: every `TransformerControlObjective` other than FIXED /
```
acostarelli: Unnecessary header.

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+# UNDEFINED. Control is opted into per `DeviceModel` with the `enable_controls` attribute;
+# when it is on, each `TransformerCircuit`'s `control_objective` decides what is built.
+# A controlled circuit gets a `TapRatioVariable` bounded by `control_limits`, and the
+# controlled quantity is held inside `controlled_quantity_limits`.
+#
+# Fixed / off-nominal tap physics (tap as a constant component property) lives in
+# `test_transformer_fixed_tap.jl` — do not duplicate it here.
```
acostarelli: The fixed tap tests are short so I don't mind putting them here. I might've already moved them.

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+#########################################################################################
+
+const VOLTAGE_CONTROL = PSY.TransformerControlObjective.VOLTAGE
+const Q_FLOW_CONTROL = PSY.TransformerControlObjective.REACTIVE_POWER_FLOW
+const P_FLOW_CONTROL = PSY.TransformerControlObjective.ACTIVE_POWER_FLOW
```
acostarelli: We're not testing active control here.

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+
+_control_attributes(enable::Bool) =
```
acostarelli: Why is this a helper

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+    Dict{String, Any}(POM.ENABLE_CONTROLS_KEY => enable)
+
+"""
+`c_sys14` with one transformer circuit put under `objective`. `regulated` picks which end
+of the circuit's arc is regulated (the bus number, not a sentinel — the API takes the
```
acostarelli: Incorrect, the API accepts any valid bus number.

acostarelli: That said we should probably include some test cases of controlling non-terminal buses. At least to make sure they are not reduced.

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+number of either the from or the to bus). Returns the system, the transformer, its
+circuit, and the regulated bus name.
+"""
+function _controlled_sys14(
+    objective;
+    name = "Trans1",
+    regulated = :to,
+    # c_sys14 buses carry (0.94, 1.06) voltage limits; a VOLTAGE band has to sit inside
```
acostarelli: Drop this comment, will be handled upstream in PSY.

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+        DeviceModel(
+            PSY.TwoWindingTransformer,
+            formulation;
+            attributes = _control_attributes(enable),
+        ),
+    )
+    return template
+end
+
+function _build_controlled(
```
acostarelli: Not sure we need three helpers here.

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+    status = build!(model; output_dir = mktempdir(; cleanup = true))
+    return model, status
+end
+
+_has_tap_variable(container) =
+    any(k -> occursin("TapRatioVariable", string(k)), keys(IOM.get_variables(container)))
+
+################################### attribute plumbing #################################
+
+@testset "enable_controls is a transformer-only attribute defaulting to false" begin
```
acostarelli: Unnecessary test

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+    sys, _, _, _ = _controlled_sys14(VOLTAGE_CONTROL; control_limits = limits)
+    model, status = _build_controlled(sys, ACPNetworkModel; optimizer = ipopt_optimizer)
+    @test status == IOM.ModelBuildStatus.BUILT
+
+    container = IOM.get_optimization_container(model)
+    tap = IOM.get_variable(container, TapRatioVariable, PSY.TwoWindingTransformer)
+    # Trans2 / Trans3 are left UNDEFINED, so only the controlled circuit gets a variable.
+    @test axes(tap)[1] == ["Trans1"]
+    @test check_variable_bounded(model, TapRatioVariable, PSY.TwoWindingTransformer)
+    for v in tap
```
acostarelli: Unnecessary tests

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+@testset "TapRatioVariable is created only for controlled circuits, bounded by control_limits" begin
+    limits = (min = 0.95, max = 1.05)
+    sys, _, _, _ = _controlled_sys14(VOLTAGE_CONTROL; control_limits = limits)
+    model, status = _build_controlled(sys, ACPNetworkModel; optimizer = ipopt_optimizer)
+    @test status == IOM.ModelBuildStatus.BUILT
+
+    container = IOM.get_optimization_container(model)
+    tap = IOM.get_variable(container, TapRatioVariable, PSY.TwoWindingTransformer)
+    # Trans2 / Trans3 are left UNDEFINED, so only the controlled circuit gets a variable.
+    @test axes(tap)[1] == ["Trans1"]
```
acostarelli: We should use `only` if we want to check that there are not other axes.

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+    @test check_variable_bounded(model, TapRatioVariable, PSY.TwoWindingTransformer)
+    for v in tap
+        @test JuMP.lower_bound(v) == limits.min
+        @test JuMP.upper_bound(v) == limits.max
+    end
+end
+
+@testset "REACTIVE_POWER_FLOW control also creates a tap variable" begin
```
acostarelli: This doesn't need to be a separate testset. We can merge with the above.

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+@testset "ACTIVE_POWER_FLOW is a phase-shift objective, not a tap control" begin
+    # The tap controls cover the voltage / reactive-power objectives; an active-power
+    # (phase-shifting) circuit must not silently acquire a tap variable.
+    sys, _, _, _ = _controlled_sys14(P_FLOW_CONTROL)
+    model, status = _build_controlled(sys, ACPNetworkModel; optimizer = ipopt_optimizer)
+    @test status == IOM.ModelBuildStatus.BUILT
+    @test !_has_tap_variable(IOM.get_optimization_container(model))
+end
+
+@testset "a DISABLED objective builds no control" begin
```
acostarelli: Drop this testset. I have no included support for these controls yet because I don't know how they work.

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+    model, status = _build_controlled(sys, ACPNetworkModel; optimizer = ipopt_optimizer)
+    @test status == IOM.ModelBuildStatus.BUILT
+    @test !_has_tap_variable(IOM.get_optimization_container(model))
+end
+
+################################### VOLTAGE objective ##################################
+
+# Solve `c_sys14` once with the transformer uncontrolled and report the regulated bus
+# voltage, so each control test can aim its band away from the free-running solution and
+# prove the constraint actually bites.
```
acostarelli: Yucky AI words

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+            optimizer = ipopt_optimizer,
+        )
+    @test status == IOM.ModelBuildStatus.BUILT
+    @test solve!(model) == IOM.RunStatus.SUCCESSFULLY_FINALIZED
+    res = IOM.OptimizationProblemOutputs(model)
+    vm = read_variable(res, "VoltageMagnitude__ACBus"; table_format = TableFormat.WIDE)
+    return vm[1, bus_name]
+end
+
+@testset "VOLTAGE control holds the regulated bus inside its band (ACP, to-side)" begin
```
acostarelli: Maybe this testset can iterate several regulated buses.

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+    tap = read_variable(
+        res, "TapRatioVariable__TwoWindingTransformer"; table_format = TableFormat.WIDE,
+    )
+    for r in 1:nrow(tap)
+        @test tap[r, "Trans1"] >= 0.9 - 1e-6
+        @test tap[r, "Trans1"] <= 1.1 + 1e-6
+    end
+end
+
+@testset "VOLTAGE control regulates the from-side bus when its number is given" begin
```
acostarelli: See comment above, we can merge this test.

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+@testset "VOLTAGE control holds the regulated bus inside its band (ACP, to-side)" begin
+    _, _, _, bus_name = _controlled_sys14(VOLTAGE_CONTROL)
+    free_vm = _uncontrolled_voltage(bus_name)
+    # A band the free-running solution violates, so holding it requires the tap to move.
+    # It sits below `free_vm`: the free-running voltage rides near the bus's 1.06 upper
+    # limit, and the band may not reach outside the bus's own limits.
+    band = (min = free_vm - 0.02, max = free_vm - 0.01)
+
+    sys, _, _, _ = _controlled_sys14(VOLTAGE_CONTROL; quantity_limits = band)
+    model, status = _build_controlled(sys, ACPNetworkModel; optimizer = ipopt_optimizer)
```
acostarelli: We should be checking all angle networks here

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+
+    res = IOM.OptimizationProblemOutputs(model)
+    vm = read_variable(res, "VoltageMagnitude__ACBus"; table_format = TableFormat.WIDE)
+    @test bus_name in names(vm)
+    for r in 1:nrow(vm)
+        @test vm[r, bus_name] >= band.min - 1e-6
+        @test vm[r, bus_name] <= band.max + 1e-6
+    end
+    # The band is on the voltage itself, not on its square.
+    @test !(free_vm >= band.min - 1e-6 && free_vm <= band.max + 1e-6)
```
acostarelli: What?

acostarelli: Brother you just made free_vm

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+
+    tap = read_variable(
+        res, "TapRatioVariable__TwoWindingTransformer"; table_format = TableFormat.WIDE,
+    )
+    for r in 1:nrow(tap)
```
acostarelli: Why are we checking tap here?

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+    vm = read_variable(res, "VoltageMagnitude__ACBus"; table_format = TableFormat.WIDE)
+    for r in 1:nrow(vm)
+        @test vm[r, bus_name] >= band.min - 1e-6
+        @test vm[r, bus_name] <= band.max + 1e-6
+    end
+end
+
+############################ REACTIVE_POWER_FLOW objective #############################
+
+@testset "REACTIVE_POWER_FLOW control holds the terminal flow inside its band (ACP)" begin
```
acostarelli: Why only ACP?

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+    # `controlled_quantity_limits` reaches the constraint builder unconverted, so it is
+    # read as system-base pu here; the reported flow is MVAR and divided back down.
+    band = (min = -0.05, max = 0.05)
```
acostarelli: Why is Claude using band when we have been using limits and bounds for everything.

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+
+    res = IOM.OptimizationProblemOutputs(model)
+    base = IOM.get_model_base_power(res)
+    for key in (
+        "FlowReactivePowerFromToVariable__TwoWindingTransformer",
+        "FlowReactivePowerToFromVariable__TwoWindingTransformer",
+    )
+        flow = read_variable(res, key; table_format = TableFormat.WIDE)
+        for r in 1:nrow(flow)
+            @test flow[r, "Trans1"] / base >= band.min - 1e-6
```
acostarelli: What base is the constraint being set in?

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+            @test flow[r, "Trans1"] / base <= band.max + 1e-6
+        end
+    end
+end
+
+################################### model invariants ###################################
+
+@testset "a tap pinned at nominal reproduces the uncontrolled model" begin
+    # White-box reduction gate: with the tap variable fixed at the circuit's nominal
```
acostarelli: Unnecessary comment, I think the testset name explains it well.

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+    # fixed-tap one, so both models must reach the same optimum and terminal flows.
+    # The band is the regulated bus's own (0.94, 1.06) limits — the widest a VOLTAGE band
+    # may be — so the control constrains nothing the bus does not already.
+    #
+    # IVR is the sharpest case: its law is multiplied through by the ratio, so every
+    # tm-bearing term has to reduce exactly for the flows to match.
+    band = (min = 0.94, max = 1.06)
+    tap_range = (min = 0.5, max = 1.5)
+
+    for network_formulation in (
```
acostarelli: The DC models also have variable tap for StaticBranchBounds.

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+            VOLTAGE_CONTROL; quantity_limits = band, control_limits = tap_range,
+        )
+        model_var, status_var = _build_controlled(
+            sys_var, network_formulation; optimizer = ipopt_optimizer,
+        )
+        @test status_var == IOM.ModelBuildStatus.BUILT
+
+        container = IOM.get_optimization_container(model_var)
+        tap = IOM.get_variable(container, TapRatioVariable, PSY.TwoWindingTransformer)
+        for t in axes(tap, 2)
```
acostarelli: We have get_time_steps

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+end
+
+################################### VOLTAGE objective ##################################
+
+# Solve `c_sys14` once with the transformer uncontrolled and report the regulated bus
+# voltage, so each control test can aim its band away from the free-running solution and
+# prove the constraint actually bites.
+function _uncontrolled_voltage(bus_name; network_formulation = ACPNetworkModel)
+    sys = PSB.build_system(PSITestSystems, "c_sys14")
+    model, status =
```
acostarelli: Weird to be calling _build_controlled in a method called _uncontrolled_voltage. I'm guessing it's just convenient but I'd rather not, it's short enough to build a model already.

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+            flow_var = read_variable(res_var, key; table_format = TableFormat.WIDE)
+            for d in PSY.get_components(PSY.TwoWindingTransformer, sys_var)
+                name = PSY.get_name(d)
+                @test isapprox(flow_var[1, name], flow_fixed[1, name]; atol = 1e-3)
+            end
+        end
+    end
+end
+
+@testset "NetworkFlowConstraint carries the live tap variable (ACP coefficients)" begin
```
acostarelli: I'm confused what this is testing. That the variable ends up in the constraint? Surely we can test this better... I feel like the controlled vs uncontrolled voltage test above does this because how else are we controlling the voltage if not through the tap? If not, then we can do a similar comparison test to compare the flows/powers.

But maybe we should also have a test that sees how a regulated bus is affected by a tap-controlled transformer whose tap we manually fix. I would hope it would be infeasible but maybe not it's just one constraint.

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+    @test isapprox(
+        JuMP.value(lookup, JuMP.constraint_object(con_pft[name, t]).func),
+        vals[pft[name, t]] - rhs;
+        atol = 1e-10,
+    )
+end
+
+################################### network coverage ###################################
+
+@testset "VOLTAGE control is wired on every voltage-carrying AC network" begin
```
acostarelli: This can be covered by tests above if we iterate through networks.

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+        # One controlled circuit, two rows (lower/upper) per time step.
+        @test length(
+            IOM.get_constraint(
+                container, VoltageMagnitudeConstraint, PSY.TwoWindingTransformer,
+            ),
+        ) == 2 * length(IOM.get_time_steps(container))
+    end
+end
+
+@testset "REACTIVE_POWER_FLOW control holds the terminal flow inside its band (IVR)" begin
```
acostarelli: Why can't we check IVR in the above test set that checks this?

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+        for r in 1:nrow(flow)
+            @test flow[r, "Trans1"] / base >= band.min - 1e-6
+            @test flow[r, "Trans1"] / base <= band.max + 1e-6
+        end
+    end
+end
+
+################################### DC networks ########################################
+
+@testset "DC networks build a variable tap under StaticBranchBounds" begin
```
acostarelli: Again, we can check this above.

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+        @test check_variable_bounded(model, TapRatioVariable, PSY.TwoWindingTransformer)
+        # Neither quantity exists on a DC network, so no band is built either way.
+        @test !IOM.has_container_key(
+            container, VoltageMagnitudeConstraint, PSY.TwoWindingTransformer,
+        )
+        @test solve!(model) == IOM.RunStatus.SUCCESSFULLY_FINALIZED
+    end
+end
+
+@testset "the DC tap law reduces to the fixed-tap law at the nominal ratio" begin
```
acostarelli: Again, our AC test does this.

acostarelli: so combine

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+
+    angle = vals[va[fr, t]] - vals[va[to, t]] - shift
+    @test isapprox(
+        JuMP.value(lookup, JuMP.constraint_object(cons[name, t]).func),
+        nominal * (vals[p[name, t]] - b_dc * angle);
+        atol = 1e-10,
+    )
+end
+
+################################ reductions and conflicts ##############################
```
acostarelli: These might better belong in a reductions test file.

acostarelli (commented):
Tests review

Have not checked that old tap tests that should be ported were ported.

test/test_voltage_control_tap_models.jl
```diff
@@ -1,405 +0,0 @@
-@testset "VoltageControlTap tap bounds are finite (Principle 0)" begin
```
acostarelli, line 1: I think this should be covered by PSY validation.

test/test_voltage_control_tap_models.jl
```diff
@@ -1,405 +0,0 @@
...
-    for tr in PSY.get_components(PSY.TwoWindingTransformer, sys)
-        lim = POM._tap_ratio_limits(tr)
-        @test isfinite(lim.min)
-        @test isfinite(lim.max)
-        @test lim.min > 0.0
-        @test lim.max >= lim.min
-    end
-end
-
-@testset "VoltageControlTap VOLTAGE objective pins regulated bus voltage (c_sys14)" begin
```
acostarelli, line 12: I think we are checking this with the comparison test, and also we don't pin at exact values anymore.

test/test_voltage_control_tap_models.jl
```diff
@@ -1,405 +0,0 @@
...
-    @test regulated_bus in names(vm)
-    for r in 1:nrow(vm)
-        @test isapprox(vm[r, regulated_bus], setpoint; atol = 1e-6)
-    end
-
-    # The tap floats within its bounds to hold the setpoint.
-    @test check_variable_bounded(model, TapRatioVariable, PSY.TwoWindingTransformer)
-end
-
-@testset "VoltageControlTap is count-invariant across control objectives (c_sys14)" begin
```
acostarelli, line 41: I could we could test this? That voltage and reactive controls make the same tap. But it feels too basic.

test/test_voltage_control_tap_models.jl
```diff
@@ -1,405 +0,0 @@
...
-    con_q = IOM.get_constraints(cq)
-    con_p = IOM.get_constraints(cp)
-    @test any(k -> occursin("RegulatedVoltageMagnitudeConstraint", string(k)), keys(con_v))
-    @test Set(keys(con_v)) == Set(keys(con_q)) == Set(keys(con_p))
-    for k in keys(con_v)
-        @test size(con_v[k]) == size(con_q[k]) == size(con_p[k])
-    end
-end
-
-@testset "VoltageControlTap @info-drop under DCPNetworkModel (c_sys14)" begin
```
acostarelli, line 265: We now support this on BranchBounds.

test/test_voltage_control_tap_models.jl
```diff
@@ -1,405 +0,0 @@
...
-    # with an @info from the (active-power-only) DC template during validation.
-    # A TwoWindingTransformer is a branch the DC network still requires to be modeled, so
-    # the build then fails on the now-unmodeled branch (unlike a droppable shunt
-    # injection). Both facts are asserted: the drop happened, and the build failed.
-    @test build!(model; output_dir = mktempdir(; cleanup = true)) ==
-          IOM.ModelBuildStatus.FAILED
-    @test !haskey(get_branch_models(get_template(model)), :TwoWindingTransformer)
-end
-
-@testset "ACP rejects two voltage regulators on one bus" begin
```
acostarelli, line 281: I think this should be a PSY validation.

test/test_voltage_control_tap_models.jl
```diff
@@ -1,405 +0,0 @@
...
-        PSY.set_regulated_bus_number!(tr, 9)
-        PSY.set_voltage_setpoint!(tr, 1.0)
-    end
-    template = get_thermal_dispatch_template_network(NetworkModel(ACPNetworkModel))
-    set_device_model!(template, PSY.TwoWindingTransformer, VoltageControlTap)
-    model = DecisionModel(template, sys; optimizer = ipopt_optimizer)
-    @test_throws IS.ConflictingInputsError POM.validate_template(model)
-end
-
-@testset "ACR does not validation-reject two regulators on one bus" begin
```
acostarelli, line 299: I'm not sure why we would want this to be the case. I think ideally we should catch infeasibilities before they appear.

test/test_voltage_control_tap_models.jl
```diff
@@ -1,405 +0,0 @@
...
-    set_device_model!(template, PSY.TwoWindingTransformer, VoltageControlTap)
-    model = DecisionModel(template, sys; optimizer = ipopt_optimizer)
-    @test POM.validate_template(model) === nothing
-end
-
-# The `_tap_flow_coefficients` hand-computed ground truth lives in
-# `test_native_transformer_tap.jl`: it exercises only `AC_branches.jl`, which is included,
-# so it must not sit behind this file's `DISABLED_TESTS` entry.
-
-@testset "ACR NetworkFlowConstraint coefficients equal _tap_flow_coefficients" begin
```
acostarelli, line 320: I see. So I think this is the test I'm confused about above. We have a round-tripping test that compares _tapped_admittance with PNM values. I can make sure that that test is okay. But I think that would exclude the need for comparing against "hand calculated" values.

test/test_native_tapcontrol.jl
```diff
@@ -1,72 +0,0 @@
...
-# `TapControl` coverage. Disabled in `runtests.jl` until
-# `ac_transmission_models/transformer_models.jl` is included again and the `TapControl`
-# formulation exists.
-#
-# The tap physics that DOES ship today — `StaticBranch` under DCP, whose susceptance is
-# tap-divided (`b_dc = 1/(tap*x)`) — is covered by `test_native_transformer_tap.jl`, which
-# runs. Do not duplicate it here.
-#########################################################################################
-
-@testset "TapControl models transformer tap ratio under DCP (c_sys14)" begin
```
acostarelli, line 11: I don't think this is a great test. I would rather test that a value that POM computes maps to a PNM-computed value in the expected way. This is just checking if a POM-computed value is computed the way a POM-computed value should be.

test/test_native_tapcontrol.jl
```diff
@@ -1,72 +0,0 @@
...
-            p_pu = flow[r, name] / base
-            expected = (va[r, fr] - va[r, to] - adm.shift) / (x * adm.tap)
-            @test isapprox(p_pu, expected; atol = 1e-5)
-        end
-    end
-    # Guard: the test system must actually have a non-unit tap, else it proves nothing.
-    @test tested_a_real_tap
-end
-
-@testset "TapControl differs from StaticBranch for non-unit-tap transformers (c_sys14)" begin
```
acostarelli, line 48: I guess this is already covered?

test/test_native_transformer_tap.jl
acostarelli: Coverage looking fine. These tests seems to be AI tests so I'm not too concerned about if these are ported.

test/test_native_transformer_tap.jl
```diff
@@ -1,114 +0,0 @@
...
-# These testsets cover only code that ships in the current module: the DC susceptance
-# `b_dc = 1/(tap*x)` used by `BThetaBranchFlow`/`NetworkFlowConstraint`, and the tap-free
-# π coefficients in `_tap_flow_coefficients` (both in `ac_transmission_models/
-# AC_branches.jl`). They deliberately do NOT touch `TapControl` / `VoltageControlTap`,
-# whose formulation files are not yet included — those live in
-# `test_native_tapcontrol.jl` / `test_voltage_control_tap_models.jl` and stay disabled
-# until the formulations are re-enabled.
-#########################################################################################
-
-@testset "StaticBranch models transformer off-nominal tap under DCP (c_sys14)" begin
```
acostarelli, line 13: We are covering this with fixed non-unit tap tests and (soon) variable tap on DC

acostarelli (commented):
Test port check. Seems fine.

test/test_native_dcp_acp_models.jl
acostarelli: Actually I'm guessing this doesn't matter.

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,755 @@
...
+@testset "REACTIVE_POWER_FLOW control also creates a tap variable" begin
+    sys, _, _, _ = _controlled_sys14(Q_FLOW_CONTROL)
+    model, status = _build_controlled(sys, ACPNetworkModel; optimizer = ipopt_optimizer)
+    @test status == IOM.ModelBuildStatus.BUILT
+    container = IOM.get_optimization_container(model)
+    @test axes(IOM.get_variable(container, TapRatioVariable, PSY.TwoWindingTransformer))[1] ==
+          ["Trans1"]
+end
+
+@testset "ACTIVE_POWER_FLOW is a phase-shift objective, not a tap control" begin
```
acostarelli: I guess. Feels similarly to the test above that no variables are created if control is off. Which is not the most useful test.

test/test_transformer_controls.jl
acostarelli: I'm actually not so sure about this.

test/test_transformer_controls.jl
acostarelli: Seems like it was Claude attempting to copy over old Claude tests.

acostarelli:
Reminder: ask Claude to reorganize all of these helpers.

src/ac_transmission_models/AC_branches.jl
```diff
@@ -69,6 +65,43 @@ function get_default_time_series_names(
...
+_control_enabled(m::DeviceModel{<:_TRANSFORMERS}) =
+    get_attribute(m, ENABLE_CONTROLS_KEY) === true
+_control_enabled(c::PSY.TransformerCircuit) =
+    PSY.get_available(c) && !(
+        PSY.get_control_objective(c) in
+        (PSY.TransformerControlObjective.UNDEFINED, PSY.TransformerControlObjective.FIXED)
+    )
+_control_enabled(_) = false
+
+_tap_controlled(c::PSY.TransformerControlObjective) = c in (
```
acostarelli: Not sure how many of these helpers are needed anymore. I wrote them before the RepresentativeBranch refactor

test/test_power_flow_in_the_loop.jl
```diff
@@ -184,72 +184,12 @@ end
 # -----------------------------------------------------------------------------
```
acostarelli, line 184: I have PFs disabled right now because there was some precompilation error. Haven't looked into it yet. It seemed like something upstream.

test/test_transformer_controls.jl
```diff
@@ -0,0 +1,881 @@
...
+end
+
+_t3w_adjacent_bus_numbers(_) = [T3W_TERMINALS..., T3W_STAR_NUMBER]
+
+#################################### case descriptors ##################################
+
+# One entry per transformer arity. Every testset below runs the whole tuple, so the two
+# arities stay in lockstep; `circuit_indices` selects which circuit of the device carries
+# the control objective (the transformer for two-winding, the winding for three-winding).
+const TWO_WINDING_CASE = (
```
acostarelli: This is kinda verbose, but it works. I also haven't looked into the 3W-specific tests too closely but I think it mostly just reused the 2W tests.

src/ac_transmission_models/RepresentativeBranch.jl
```diff
@@ -0,0 +1,284 @@
...
+    reduction::String
+    branch::B
+    nr::PNM.NetworkReductionData
+    number_to_name::Dict{Int, String}
+end
+
+"""
+Used for specializing the device loop per concrete RepresentativeBranch.
+"""
+function _for_each_branch(f::F, reps) where {F}
```
acostarelli: Closures are unusual for Sienna, but we previously had the issue where the type of the branches we were iterating could've been Lines, MonitoredLines, TwoWindingTransformers, ThreeWindingTransformerCircuits, or some PNM.AbstractReductionAggregate. We need to guard the inner loop by a function so it can specialize.

That said, this is a pretty bare-bones method. Maybe that's okay. Maybe we combine to make _foreach_representative_branch and _foreach_branch ? The issue is a lot of call-sites need the branches vector before iterating so that they can build the names axis for a container, which is an annoying pattern I've encountered several times -- I wish we didn't have to initialize containers with names.

test/includes.jl
```diff
@@ -7,7 +7,7 @@ using InfrastructureSystems
 import InfrastructureSystems: TableFormat
 using PowerNetworkMatrices
 import PowerSystemCaseBuilder: PSITestSystems
-using PowerFlows
+#using PowerFlows
```
m-bossart: We shouldn't merge this with PowerFlows removed from the tests entirely. My understanding is that psy6 in PowerFlows should be compatible.

test/test_ac_transmission_security_constrained_models.jl
```diff
@@ -1564,17 +1564,6 @@ end
     @test lim.min ≈ -rb
 end
 
-@testset "emergency limits: TwoWindingTransformer rating_b resolves through its circuit" begin
```
m-bossart, line 1567: Why is this test deleted?

test/test_device_branch_constructors.jl
```diff
@@ -1017,10 +1017,12 @@ end
 # PSY change introducing a per-branch base surfaces here instead of silently mis-bounding
 # branch flows against the system-base `FlowActivePowerVariable` bounds.
 @testset "PNM rating aggregators are system base (branch_rating invariant)" begin
+    _rating(t::PSY.TwoWindingTransformer) = PSY.get_rating(PSY.get_circuit(t), PSY.SU)
```
m-bossart: If you are going to defined this function, might as well use it above instead of repeating the pattern.

test/test_native_dcp_acp_models.jl
```diff
@@ -483,29 +483,20 @@ end
...
-    #         get_thermal_dispatch_template_network(NetworkModel(network_formulation))
-    #     set_device_model!(
-    #         template,
-    #         DeviceModel(PSY.Line, StaticBranchUnbounded; use_slacks = true),
-    #     )
-    #     model = DecisionModel(template, sys; optimizer = HiGHS_optimizer)
-    #     @test_throws IS.ConflictingInputsError POM.validate_template(model)
-    # end
-
-    # sys14 = PSB.build_system(PSITestSystems, "c_sys14")
```
m-bossart, line 501: Why is this test deleted? The previous comment said that the Voltage control never adds slacks.

test/test_native_network_reductions.jl
```diff
@@ -570,7 +613,8 @@ end
     device_model = get_model(get_template(model), PSY.Line)
     for (name, (arc, reduction)) in line_entries
         entry = all_maps[reduction][PSY.Line][arc]
-        rating = POM.branch_rating(entry, device_model)
+        rating = POM._branch_rating(entry, device_model)
+        rating = _
```
m-bossart: ?

test/test_native_network_reductions.jl
```diff
@@ -411,16 +411,59 @@ end
...
+        @test axes(
+            IOM.get_variable(container, TapRatioVariable, PSY.TwoWindingTransformer),
+        )[1] ==
+              ["Trans$i"]
+    end
+end
+
+@testset "a controlled circuit merged with a parallel branch fails with a clear error" begin
+    sys, transformer, circuit, _ = _controlled_sys14(VOLTAGE_CONTROL)
+    arc = PSY.get_arc(circuit)
```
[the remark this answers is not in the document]
acostarelli: This does not cause an error, it just means we miss the last value. But since we aren't taking the last value(s) anyway, it's okay.

src/network_models/reduction_exceptions.jl
```diff
@@ -183,3 +154,26 @@ function _pin_model_all_branches!(
...
+function _pin_transformer_controls!(
+    buses::Set{Int},
+    m::DeviceModel{<:_TRANSFORMERS},
+)
+    _control_enabled(m) || return
+    for transformer in get_device_cache(m)
+        for circuit in PSY.get_circuits(transformer)
+            PSY.get_available(circuit) || continue
+            PSY.get_control_objective(circuit) in
+            (PSY.TransformerControlObjective.VOLTAGE,) || continue
```
m-bossart: Shouldn't we be pinning buses when we have reactive power control also?

m-bossart (commented):
A few comments, we can discuss more today

test/test_ac_transmission_security_constrained_models.jl
acostarelli: It seems a little pointless.

test/test_native_dcp_acp_models.jl
acostarelli: We don't have this formulation anymore.

src/network_models/reduction_exceptions.jl
acostarelli: Yes this was supposed to be one more line down.

acostarelli:
@m-bossart I addressed your reviews. I have some responses to the one from Slack:
7. Should reactive power control only constrain the from-to flow? This is what I changed it to. Previously it was constraining both.
9. I added validation that the regulated bus number is valid, but I think this should happen in PSY
12. I'm choosing not to use regulated_voltage.jl because it feels quite overkill and my solution is cleaner IMO.
13. Rather than restoring the test with "hand-computed" susceptances, I have the test compare POM's calculated admittance against a DC PF, to ensure tap is included correctly

test/test_power_flow_in_the_loop.jl
```diff
@@ -184,72 +184,12 @@ end
...
 # Baseline PFitL coverage (ported from PowerSimulations.jl test file lines 1-548).
 # These exercise the regular non-headroom paths through the migrated code:
-#   - PhaseShiftingTransformer in PFitL
 #   - Parallel-line aggregation
 #   - Breaker-switch (DiscreteControlledACBranch)
 #   - HVDCs with DC PowerFlow
 #   - Line active power loss aux variable
 # -----------------------------------------------------------------------------
 
-@testset "AC Power Flow in the loop for PhaseShiftingTransformer" begin
```
m-bossart, line 194: Why is this test deleted? I think this should be left commented and restored when the phase shifting formulation is completed.

src/ac_transmission_models/RepresentativeBranch.jl
```diff
@@ -0,0 +1,315 @@
...
+    PSY.get_rating(t.circuit, PSY.SU)
+_branch_rating(entry::PNM.BranchesSeries, ::DeviceModel) = PNM.get_equivalent_rating(entry)
+_branch_rating(entry::PNM.AbstractBranchesParallel, model::DeviceModel) =
+    _parallel_branches_rating(model, entry)
+
+_branch_rating(rep::RepresentativeBranch, model::DeviceModel) =
+    _branch_rating(rep.branch, model)
+
+# TODO: SC branches don't use the RepBranch API but they need rating_b, so
+# move this or change SC to use this API.
```
m-bossart: `_branch_rating_b` is dispatching to just `get_rating`. This will return the normal rating even when there is an emergency rating defined. Something wrong here.

src/ac_transmission_models/RepresentativeBranch.jl
```diff
@@ -0,0 +1,315 @@
...
+_branch_rating(t::PNM.ThreeWindingTransformerCircuit, ::DeviceModel) =
+    PSY.get_rating(t.circuit, PSY.SU)
+_branch_rating(entry::PNM.BranchesSeries, ::DeviceModel) = PNM.get_equivalent_rating(entry)
+_branch_rating(entry::PNM.AbstractBranchesParallel, model::DeviceModel) =
+    _parallel_branches_rating(model, entry)
+
+_branch_rating(rep::RepresentativeBranch, model::DeviceModel) =
+    _branch_rating(rep.branch, model)
+
+# TODO: SC branches don't use the RepBranch API but they need rating_b, so
```
m-bossart: Can we address this now and not leave a TODO? What are the limitations to using RepresentativeBranch with SC models?

src/ac_transmission_models/AC_branches.jl
```diff
@@ -128,155 +114,215 @@ function get_default_attributes(
...
+    branch,
+    device_model::DeviceModel,
+    network_model::NetworkModel,
+) = _tap_controlled(branch, device_model, network_model)
+_branch_uses_control(
+    ::Type{PhaseShifterAngle},
+    branch,
+    device_model::DeviceModel,
+    network_model::NetworkModel,
+) = _phase_controlled(branch, device_model, network_model)
```
m-bossart: _phase_controlled is not defined. Probably OK for now given that this will be fixed in the follow up work implementing phase shifting, but this should be probably commented with a TODO instead.

src/ac_transmission_models/AC_branches.jl
```diff
@@ -297,116 +343,14 @@ function _add_meta_flow_slack!(
...
-    return get_min_max_limits(device, FlowRateConstraint, AbstractBranchFormulation)
-end
-
 # Branch-rating time-series multiplier at build time. Non-parallel entries use
 # the same aggregation as the static `branch_rating` path. Parallel groups are
 # the exception: a series on one member can't be split across the group, so the
 # summed (emergency) rating is used regardless of the attribute. Every PNM
 # reduction wrapper is `<: PSY.ACTransmission`; the parallel methods are more
-# specific (`<: AbstractBranchesParallel`), so they win for groups.
+# specific (`<: AbstractBranchesParallel`), so they win for reproups.
```
m-bossart: typo

src/ac_transmission_models/AC_branches.jl
```diff
@@ -69,6 +42,17 @@ function get_default_time_series_names(
     return Dict{Type{<:TimeSeriesParameter}, String}()
 end
 
+const _TRANSFORMERS = Union{PSY.TwoWindingTransformer, PSY.ThreeWindingTransformer}
+
+const ENABLE_CONTROLS_KEY = "enable_controls"
+
+_control_attribute(::Union{Type{<:_TRANSFORMERS}}) = (ENABLE_CONTROLS_KEY => false,)
```
m-bossart: Single element Union doesn't make sense

src/ac_transmission_models/branch_constructor.jl
```diff
@@ -507,6 +511,7 @@ ArgumentConstructStage for StaticBranch under LPACCNetworkModel.
 Creates the four directional flow variables, the bus-pair cosine variable (cs), optional
 slacks, and registers each flow's contribution to the per-bus ActivePowerBalance and
 ReactivePowerBalance expressions.
+
```
m-bossart: remove

m-bossart (commented):
A few general comments. 
1. The rating_b thing is the most concerning, I think that is a real regression that is made in this PR. It might be that we need better test coverage there. 
2. There is still a bunch of commented/stale code that is no longer needed after this refactor. Search for `"psy6: disabled pending transformer refactor"`. We can leave some of the code related to phase shifting for now, but some of that can be cleaned up.
3. Update the documentation on the transformer formulations. All of the information there is now stale. 
4. `TapRatioVariable`'s exported docstring
  (core/variables.jl:65-66) reads "for a the disabled VoltageControlTap formulation". It references a deleted type on the public docstring. 
5. I think it is OK to not reuse regulated_voltage.jl because we are implementing a voltage band and not fixing a specific value. But we should do some cleanup to that file to remove the functionality that was originally included for transformers (including the documentation).

src/ac_transmission_models/RepresentativeBranch.jl
acostarelli: Typo

acostarelli: But yes coverage is not great

src/ac_transmission_models/RepresentativeBranch.jl
acostarelli: Seems like that whole _check_time_series method in add_parameters.jl isn't covered because it doesn't even have a correct _branch_rating signature.

jd-lara (approved):

PR #243, opened 2026-08-19
Move `get_hinted_aff_expr` from IOM into POM

### LINKED ISSUE #224: Move `get_hinted_aff_expr` out of IOM (unexported, and IOM never calls it)
Raised by @luke-kiernan reviewing #206:

> organizational nitpick: `get_hinted_aff_expr` is non-exported, but the only places that call it are in POM. Move to POM or export it

Deferred out of #206 because that PR should not touch IOM. Filing so it does not get lost.

## Verified state

`get_hinted_aff_expr` is defined in IOM at `src/utils/jump_utils.jl:10`.
It is three lines: `JuMP.AffExpr(0.0)` plus a `sizehint!`.

- **Not exported.** It appears in none of the ~289 `export` lines in `src/InfrastructureOptimizationModels.jl`.
- **IOM never calls it.** A full-tree grep for `hinted` across IOM `src`, `test` and `docs` returns only the definition itself. It is dead code inside its own package.
- **Every caller is POM**, six sites, all `IOM.`-qualified:
  - `src/ac_transmission_models/AC_branches.jl:776`
  - `src/ac_transmission_models/AC_branches.jl:798`
  - `src/services_models/reserves.jl:145`
  - `test/test_ac_transmission_security_constrained_models.jl:116`
  - `test/test_ac_transmission_security_constrained_models.jl:234`
  - `test/test_ac_transmission_security_constrained_models.jl:333`
- **Not in InfrastructureSystems.**
- `PowerSimulations` has its own private copy at `src/utils/jump_utils.jl:3`. It is not a consumer of IOM's, so it does not constrain the decision.

Note this is not one of the `IOM._*` underscore helpers POM reaches into.
It is a public-*looking* name that happens to be unexported, which is arguably worse for discoverability than an explicit underscore.

## Options

**Move it to POM** (preferred). Put it in POM `src/utils/`, drop the `IOM.` qualifier at the six sites, and delete it from IOM in a companion PR.
The function has no IOM caller to break, so this is a clean removal rather than a deprecation.

**Export it from IOM.** Cheaper, but it publicizes a function IOM itself has no use for, which is the wrong signal about where the abstraction belongs.

## Sequencing

Two PRs, POM first so IOM is never the broken side:

1. POM: add the definition locally, switch the six call sites, keep working against current IOM.
2. IOM: delete `get_hinted_aff_expr` from `src/utils/jump_utils.jl`.

Related to the broader concern that POM currently calls ~56 non-exported `IOM._*` helpers, an informal but load-bearing surface.
This one is a small, self-contained instance of the same boundary question.

luke-kiernan:
Fixes #224 (POM half).

`IOM.get_hinted_aff_expr` is unexported and has no caller inside IOM — every call site is in POM. This moves the three-line helper into POM at `src/utils/jump_utils.jl` and drops the `IOM.` qualifier at all seven call sites (four in `src`, three in `test`; the test sites take a `POM.` qualifier since the helper stays unexported).

A companion PR on InfrastructureOptimizationModels.jl deletes the original definition. Merging this one first keeps IOM from ever being the broken side.

No behavior change.

### DIFF
--- a/src/PowerOperationsModels.jl
+++ b/src/PowerOperationsModels.jl
@@ -232,6 +232,7 @@ import InfrastructureOptimizationModels:
 # These define concrete Variable, Expression, Constraint, and Parameter types
 # and extend should_write_resulting_value/convert_output_to_natural_units
 #################################################################################
+include("utils/jump_utils.jl")
 include("core/definitions.jl")
 include("core/problem_types.jl")
 include("core/interfaces.jl")
--- a/src/ac_transmission_models/AC_branches.jl
+++ b/src/ac_transmission_models/AC_branches.jl
@@ -774,7 +774,7 @@ function _make_flow_expressions!(
     hint = length(nz_idx)
     expressions = Vector{JuMP.AffExpr}(undef, length(time_steps))
     for t in time_steps
-        acc = IOM.get_hinted_aff_expr(hint)
+        acc = get_hinted_aff_expr(hint)
         @inbounds for i in nz_idx
             JuMP.add_to_expression!(acc, ptdf_col[i], nodal_balance_expressions[i, t])
         end
@@ -798,7 +798,7 @@ function _make_flow_expressions!(
     hint = length(nz_idx)
     expressions = Vector{JuMP.AffExpr}(undef, length(time_steps))
     for t in time_steps
-        acc = IOM.get_hinted_aff_expr(hint)
+        acc = get_hinted_aff_expr(hint)
         @inbounds for k in eachindex(nz_idx)
             JuMP.add_to_expression!(
                 acc,
--- a/src/services_models/reserve_group.jl
+++ b/src/services_models/reserve_group.jl
@@ -129,7 +129,7 @@ function add_constraints!(
 
     for t in time_steps
         vars = member_vars[t]
-        resource_expression = IOM.get_hinted_aff_expr(length(vars))
+        resource_expression = get_hinted_aff_expr(length(vars))
         for var in vars
             JuMP.add_to_expression!(resource_expression, var)
         end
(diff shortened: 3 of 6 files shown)


acostarelli (commented):
I wonder how this compares to building each expression by a comprehension in `JuMP.@expression`, performance-wise.

Separately, this move feels strange because POM doesn't have a `jump_utils.jl` and now it has a tiny one.

luke-kiernan:
> this move feels strange because POM doesn't have a jump_utils.jl and now it has a tiny one.

Fair.

> I wonder how this compares to building each expression by a comprehension in JuMP.@expression, performance-wise.

I'd expect this to be faster or the same. Comprenhensions usually end up creating intermediates under the hood, which means extra allocations. That's what we're trying to avoid here with `sizehint!(expr.terms, size)`. 

Out of curiousity I asked AI about this. When the iterators/generators have well-defined sizes (e.g. vectors), `Dict(zip(keys_itr, vals_itr))` will automatically `sizehint!` so it only allocates once. `d = Dict(k => f(k) for k in keys_vec)` is better than `d = Dict([k => f(k) for k in keys_vec])`: 2nd will allocate an intermediate vector. Julia lacks move semantics, so there is no equivalent of C++'s memory-efficient add-to-end functions like `emplace_back` and `extend`.

jd-lara:
This is relevant to PSINV. Please don't move core methods out of POM

luke-kiernan:
> This is relevant to PSINV. Please don't move core methods out of POM

I was super confused, then realized that you meant "out of IOM" _not_ "out of POM." Sure. I'm inferring that PSINV has only IOM as a dependency, not POM, which is why moving it IOM -> POM would be problematic

